PR #15, opened 2022-07-27
Building a Tiny JS World

kovalov:
### Building a Tiny JS World

[Demo](https://kovalov.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/kovalov/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/kovalov/js-pre-oop/index.js
+++ b/submissions/kovalov/js-pre-oop/index.js
@@ -1,0 +1,50 @@
+const dog = {
+  name: 'Patron',
+  gender: 'male',
+  species: 'dog',
+  legs: 4,
+  hands: 0,
+  saying: 'woof',
+};
+
+const cat = {
+  name: 'Murzyk',
+  gender: 'male',
+  species: 'cat',
+  legs: 4,
+  hands: 0,
+  saying: 'meow',
+};
+
+const woman = {
+  name: 'Anna',
+  gender: 'female',
+  species: 'human',
+  legs: 2,
+  hands: 2,
+  saying: 'hello',
+};
+
+const man = {
+  name: 'Joey',
+  gender: 'male',
+  species: 'human',
+  legs: 2,
+  hands: 2,
+  saying: 'How you doin?',
+};
+
+const catWoman = Object.create(cat);
+catWoman.name = 'Cat-woman';
+catWoman.species = 'human';
+catWoman.legs = 2;
+catWoman.hands = 2;
+
+const inhabitantsArray = [dog, cat, woman, man, catWoman];
+
+inhabitantsArray
+  .map(
+    (item) =>
+      `${item.species}; ${item.name}; ${item.gender}; ${item.legs}; ${item.hands}; ${item.saying}`
+  )
+  .forEach((message) => print(message));


submissions/kovalov/js-pre-oop/index.js
```diff
@@ -0,0 +1,50 @@
...
+  saying: 'How you doin?',
+};
+
+const catWoman = Object.create(cat);
+catWoman.name = 'Cat-woman';
+catWoman.species = 'human';
+catWoman.legs = 2;
+catWoman.hands = 2;
+
+const inhabitantsArray = [dog, cat, woman, man, catWoman];
```
artem-trubin, line 43: It is usually bad to specify data type of a value, `inhabitants` already implies plural, so array.

submissions/kovalov/js-pre-oop/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+inhabitantsArray
+  .map(
+    (item) =>
+      `${item.species}; ${item.name}; ${item.gender}; ${item.legs}; ${item.hands}; ${item.saying}`
```
artem-trubin, line 48: Can you make it even better by somehow using a high order array function?

submissions/kovalov/js-pre-oop/index.js
```diff
@@ -0,0 +1,50 @@
...
+  )
+  .forEach((message) => print(message));
```
artem-trubin, line 50: Couldn't you just skip a `map` method and go for the `forEach` straightaway?

submissions/kovalov/js-pre-oop/index.js
```diff
@@ -0,0 +1,50 @@
...
+const man = {
+  name: 'Joey',
+  gender: 'male',
+  species: 'human',
+  legs: 2,
+  hands: 2,
+  saying: 'How you doin?',
+};
+
+const catWoman = Object.create(cat);
```
artem-trubin, line 37: Is it fine that your cat-woman is a male?

artem-trubin (commented):
Nice job! Now let's make it even better together!

submissions/kovalov/js-pre-oop/index.js
kovalov: Should it look like that:
`function fn(arr, callback) {
   return function(){
         callback(arr);
      }
}`
?

kovalov: Completed

submissions/kovalov/js-pre-oop/index.js
```diff
@@ -0,0 +1,46 @@
...
+catWoman.name = 'Cat-woman';
+catWoman.species = 'human';
+catWoman.gender = 'female';
+catWoman.legs = 2;
+catWoman.hands = 2;
+
+const inhabitants = [dog, cat, woman, man, catWoman];
+
+inhabitants.forEach((item) => print(Object.values(item).join('; ')));
```
OleksiyRudenko, line 46: Well done.

Note that (a) `Object` methods doesn't guarantee a particular order of keys or values and (b) other contributors to your project may define new objects' props in a different order (and you may miss this during code review), also (c) project may require adding "private" properties to the objects, so the recommended approach is to list required properties explicitly in  use cases like this one.

It's not a hard requirement in this particular task but you may want to accept this small challenge to develop a proper habit.

OleksiyRudenko (commented):
Good job! Check the comment below to see if you want to adopt a useful pattern.

submissions/kovalov/js-pre-oop/index.js
kovalov: Updated. Please review.

submissions/kovalov/js-pre-oop/index.js
```diff
@@ -1,45 +1,52 @@
...
+const keys = ['species', 'name', 'gender', 'legs', 'hands', 'saying'];
+
+const details = inhabitants.map((item) => {
+  return keys.map((key) => item[key]);
+});
 
-inhabitants
-  .map((item) => Array.from(item.values()))
-  .forEach((item) => print(item.join('; ')));
+details.forEach((item) => print(item.join('; ')));
```
OleksiyRudenko, line 52: ### 1. Arrow functions and chaining

If an arrow's function is a one-liner then you may skip `return` keyword and curly brackets.
If arrow function has a single argument then parentheses can be omitted.

Compare
```
const details = inhabitants.map((item) => {
  return keys.map((key) => item[key]);
});

const details = inhabitants.map(item => keys.map(key => item[key]));
```
Definitely, if you have more nesting levels in your callbacks you will want to split the line into several lines for readability or even split into several independent expressions (just like you've done).
Having no extra keywords like `return` and unnecessary brackets will improve readability.
```
const details =  inhabitants.map((item) => {
  return keys.map((key) => item[key]);
});

const details = inhabitants.map(item =>
  keys.map(key => item[key])
);

// or
const details = inhabitants.map(
  item => keys.map(key => item[key])
);

// or
const details = inhabitants.map(
  item => keys.map(key => item[key]).join("; ") 
  // with `.join` each entry in `details` array represents a ready made presentation string
); // here you may chain `.forEach` and wouldn't need `details` variable
```
Linter will help you to format code for better readability but will unlikely refactor the code to have no `return` statements.

### 2. Naming

Avoid generic, unrelated to code business domain terms and words when naming variables and functions/methods, whenever you implement a specific business logic. Generics are OK when we design a multipurpose functions or library/SDK.

Imagine, your codebase is split into fragments - a method, or function, or declaration are each in their own file.

So reading a specific file must tell the reader what is going on there, provided they know general context of the task.

Variable names
- `dog`, `cat` etc - good, clear what/who an object represents;
- `inhabitants` - good, clearly understandable what to expect to see inside: a collection of world innhabitans;
- `keys` - too generic; we **know** it contains a list of an inhabitant's properties; it would be better if it **told** the reader this, e.g. **inhabitantPropertyNames**;
- `key` - too generic; what is it? an inhabitant's prop name? tell this;
- `item` - too generic; what specifically a callback receives? is it an inhabitant or an inhabitant's details? tell this;

Don't save characters or typing effort on naming. Luckily, IDE offers autoco,pletion once a variable defined.

OleksiyRudenko (approved):
Great job!

The comments below do not require any implementation. I am good.

Make sure other reviewers' comments, suggestions and questions are addressed.

kovalov:
Please review the final state of repo and comment (or merge).

artem-trubin:
May I ask you why did you change the way catwoman is created? Task states that catwoman should copy what cat says, even if we change what cat says in the process. You previous implementation allowed that, now you're only copying the initial value. Changes to cat won't transfer to catwoman now.

kovalov:
> May I ask you why did you change the way catwoman is created? Task states that catwoman should copy what cat says, even if we change what cat says in the process. You previous implementation allowed that, now you're only copying the initial value. Changes to cat won't transfer to catwoman now.

My fault. Have tried to make it simpler.
Have amended naming and and added prototype.

PR #18, opened 2022-07-27
task-dom

maximmorenko:
### task-dom

[Demo](https://maximmorenko.github.io/kottans-frontend-dom) |
[Code base](https://github.com/maximmorenko/kottans-frontend-dom)

### DIFF
--- a/submissions/morenko/dom/index.html
+++ b/submissions/morenko/dom/index.html
@@ -1,0 +1,46 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" href="src/styles.css">
+    <title>facts</title>
+    <meta charset="UTF-8" />
+  </head>
+  <body>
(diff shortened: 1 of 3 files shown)


artem-trubin:
H! Let's make some changes.

First of all, your site doesn't work well with small screens. Adaptivity is always required at this point.
image

Secondly, your buttons have strange hitboxes. I believe they only work if user clicks on text, not on the button itself. It is confusing, fix please.

Another thing that I see - you reach out to DOM methods like `querySelector` A LOT. These are expensive and if you have an element that is going to be present on the page and not deleted, you can just use `querySelector` for it once and store in in a constant. It will work like a link and you won't have to reach out to it every time you need it.

By the way, using keyboard to navigate the menu doesn't work. You know how to do this already, please implement it here.

And please check my comments before this message.

PR #29, opened 2022-07-28
Vlad dom

Pozzitive11:
### dom-js

[Demo](https://pozzitive11.github.io/dom-js/) |
[Code base](https://github.com/Pozzitive11/dom-js)

Please, review.


submissions/pozzitive11/vlad-js-dom/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+const createBtn = (name, img, id) => {
+  return `<li class ="nav__item">
+            <button class="nav__btn" id="${id}" style="background-image: url(${img});">${name}</button>
+          </li>`;
+};
+
+function createNavList(photos) {
+  let btns = photos.map(({ name, img, id }) => createBtn(name, img, id));
+  let navList = document.querySelector(".nav__list");
```
artem-trubin, line 54: `const navList` and `const btns`

submissions/pozzitive11/vlad-js-dom/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+// Create btns
+
+const createBtn = (name, img, id) => {
+  return `<li class ="nav__item">
+            <button class="nav__btn" id="${id}" style="background-image: url(${img});">${name}</button>
+          </li>`;
+};
+
+function createNavList(photos) {
```
artem-trubin, line 52: Is there a reasong for this function to exist? It seems like the only time you use it is right after you define it and that's it.

submissions/pozzitive11/vlad-js-dom/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+const fillContent = (item) => {
+  img.innerHTML = `<img class="photo-img" src="${item.img}" alt="${item.name}" />`;
+  nameImg.textContent = item.name;
+  creator.textContent = item.creator;
+  description.textContent = item.description;
+};
+fillContent(photos[0]);
+
+const navList = document.querySelector(".nav__list");
```
artem-trubin, line 74: You already got `navList` in the function `createNavList`. If you could just place `createNavList` code outside the function scope, you would already have this variable.

submissions/pozzitive11/vlad-js-dom/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+navList.addEventListener("click", (event) => {
```
artem-trubin, line 76: You are not using `event` anywhere except the next line, it makes sence to desctruct it and just get `{target}` as a function argument.

submissions/pozzitive11/vlad-js-dom/script.js
```diff
@@ -0,0 +1,93 @@
...
+  const target = event.target;
+  if (target && target.classList.contains("nav__btn")) {
+    let photoId = target.id;
+    let photoItem = photos.find((item) => item.id === photoId);
```
artem-trubin, line 80: `const`

artem-trubin (commented):
Hi! Nice work!

There is a little problem with mobile view - hamburger menu button overlaps with header (see screenshot). It would be nice to move some thing around, so they are placed better.

image

Also, check comments below.

submissions/pozzitive11/vlad-js-dom/script.js
Pozzitive11: i.e. do like this?
image

artem-trubin: What I meant is that you can recieve `{ target }` in the functions arguments, instead of `event`. This will extract `target` value from the incoming `event` argument and establish it as local value. You will not need to use `event.target` each time you need it after these changes.

Pozzitive11: done 😃

submissions/pozzitive11/vlad-js-dom/index.html
```diff
@@ -0,0 +1,55 @@
...
+          </ul>
+        </aside>
+
+        <div class="content">
+          <figure class="photo">
+            <div class="photo__img">
+              <img class="photo-img" src="img/mystical-mountain.jpg" alt="">
+            </div>
+            <figcaption>
+              <div class="photo__name"></div>
```
artem-trubin, line 39: This probably should be a `h2` tag with a page title, not a photo description.

artem-trubin:
Well done!

PR #35, opened 2022-07-30
tiny-js-world

AntonKottans:
Building a Tiny JS World

[Demo](https://antonkottans.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/AntonKottans/a-tiny-JS-world)

Please, review.


artem-trubin (commented):
Brilliant!

There is one problem, though. Task states "cat-woman should magically always tell what cat does". In your current implementation if we change cat's saying, catwoman's one will not change.

artem-trubin:
Clever thinking!

PR #36, opened 2022-07-30
a-tiny-JS-world

Eugene-CG:
### a-tiny-JS-world

[Demo](https://eugene-cg.github.io/a-tiny-JS-world/)
[Code base](https://github.com/Eugene-CG/a-tiny-JS-world)

Please, review.


submissions/Eugene-CG/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,81 @@
...
+  hands: 2,
+  saying: "Nice to meet you",
+};
+const catWoman = {
+  species: "Dream",
+  name: "Noone",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: cat.saying,
```
artem-trubin, line 49: Here you are only copying the initial value of the `cat.saying`. You have to figure out some way to connect those values, so that if you change `cat.saying`, `catWoman.saying` changes as well.

submissions/Eugene-CG/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,81 @@
...
+};
+const arr = [cat, dog, man, woman, catWoman];
```
artem-trubin, line 51: Avoid using data types in the name of the variable. It is a clear indicator that your variable name isn't as obvious as it should be. Let's take this one for example. `arr` of what? If you use this variable outside this code somewhere, would you be able to understand what data is inside? Let's replace it with something like `inhabitants`. This gives you and understanding of what data is there and the fact that this is plural suggests that this is an array as well.

submissions/Eugene-CG/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,81 @@
...
+let string = [];
```
artem-trubin, line 52: Same thing, but even more confusing. A `string` that is actually and array? Also, you could've `print`ed every inhabitants info separately, it is not against the task, but since you are already doing this, why `let`? You are not replacing the value anywhere, you are using `push`. It is allowed to change the value of arrays with functions like `push` even when the variable is a `const`.

artem-trubin (commented):
Good job! Now let's make it even better.

Eugene-CG:
catWoman.prototype = cat.prototype, but I used Object.setPrototypeOf, because Object.create(prototype, descriptors) dont work properly, I dont know why.
Object.create(getPrototypeOf(cat), ...) dont work

submissions/Eugene-CG/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,85 @@
...
+  "gender",
+  "legs",
+  "hands",
+  "saying",
+];
+inhabitants.forEach((obj) => {
+  print(
+    inhabitantsKeys.reduce((accumulator, prop) => {
+      return (accumulator += obj[prop] + "; ");
+    }, "")
```
OleksiyRudenko, line 73: Consider using `#Array.map` (to map property names into values) and `#Array.join` (to glue substrings together) here. This will make the code cleaner.

OleksiyRudenko (changes requested):
@Eugene-CG good job!
A bit of improvement and we are done.

submissions/Eugene-CG/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,84 @@
...
+  "gender",
+  "legs",
+  "hands",
+  "saying",
+];
+inhabitants.map((obj) => {
+  print(
+    inhabitantsKeys.map((prop) => {
+      return obj[prop];
+    })
```
OleksiyRudenko, line 73: This can be reduced to `inhabitantsKeys.map(prop => obj[prop])`. `return` statement and surrounding `{}` are redundant if `return` is the only statement in arrow function body.

OleksiyRudenko (approved):
@Eugene-CG well done!
Check the comment below to improve your code style a bit..

PR #39, opened 2022-07-30
add task task_js_dom

Sadoharu:
### Document Object Model

[Demo](https://sadoharu.github.io/task_js_dom/) |
[Code base](https://github.com/Sadoharu/task_js_dom)

Please, review.


submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+            <span class="nav-bar__list_title">${item.title}</span>
+            `;
+        navList.append(liItem);
+    });
+
+    const hideShowPage = (show = 0) => {
+        pages.forEach(item => {
+            item.style.display = "none";
+        });
+        pages[show].style.display = "block";
```
zonzujiro, line 29: Please don't manipulate styles directly from JS, because it is considered an antipattern.
Better will be to add class `visible` or smth

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+        liItem.tabIndex = 0;
+        liItem.setAttribute('data-index', i);
+        liItem.innerHTML = `
+            <span class="nav-bar__list_icon ${item.icon}"></span>
+            <span class="nav-bar__list_title">${item.title}</span>
+            `;
+        navList.append(liItem);
+    });
+
+    const hideShowPage = (show = 0) => {
```
zonzujiro, line 25:
```suggestion
    const togglePageVisibility = (show = 0) => {
```

zonzujiro: And what is `0` and why it's `show`? 🤔

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+        node.classList.add('active');
+    };
+
+    navList.addEventListener('click', e => {
+        if (e.target.getAttribute('data-index')) {
+            togglePage(e, e.target);
+        } else if (e.target.parentNode.getAttribute('data-index')) {
+            togglePage(e, e.target.parentNode);
+        }
+    });
```
zonzujiro, line 46: Please, apply [prettier](https://prettier.io/playground/) to the code, to whole file

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+            item.style.display = "none";
+        });
+        pages[show].style.display = "block";
+    };
+
+    hideShowPage();
+
+    const togglePage = (e, node) => {
+        hideShowPage(node.getAttribute('data-index'));
+        e.currentTarget.querySelectorAll('.nav-bar__list_item').forEach(item => item.classList.remove('active'));
```
zonzujiro, line 36: Looks like instead of changing the class name on every item, you can change class on the parent. ;)

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+        node.classList.add('active');
+    };
+
+    navList.addEventListener('click', e => {
+        if (e.target.getAttribute('data-index')) {
+            togglePage(e, e.target);
+        } else if (e.target.parentNode.getAttribute('data-index')) {
```
zonzujiro, line 43: Don't use things like `parentNode` etc. for accessing elements. Someone somewhere can add an additional wrapper between your files, which will break your code.

First of all, you have [`closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest).

Second, I believe you can just have one listener on the parent and use the benefits of event bubbling.

zonzujiro (changes requested):

zonzujiro:
Great. Amazing. Never seen anything better.

Please go thru comments :)

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: show - argument
0 - default parameter

https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Functions/Default_parameters

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: i need to give a class to a specific button. How can I give a certain button the class active while working with the parent // мені потрібно надати класс конкретній кнопці. Як я зможу певній кнопці надати класс 'active' працюючи з батьком

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: fix it

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: change to closest. thanks

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: done

Sadoharu:
> Great. Amazing. Never seen anything better.
> 
> Please go thru comments :)

good joke

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: the same with the deletion of the class

submissions/sadoharu/task_js_dom_/scripts.js
zonzujiro: I know what is arguments and I know what is `0` in this case. :)

The question is, why you are using them? Why do you need them? What is the idea behind them? Why it's a `0` and not a `boolean`?

submissions/sadoharu/task_js_dom_/scripts.js
zonzujiro: I don't think you need this logic in general :) I have a feeling, that you can just read about event bubbling and solve this with event delegation ;)

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,73 @@
...
+      {
+        title: "Reviews",
+        icon: "icon-rate_review",
+      },
+    ];
+  
+    navListItem.forEach((item, i) => {
+      const liItem = document.createElement("li");
+      liItem.classList.add("nav-bar__list_item");
+      liItem.tabIndex = 0;
```
zonzujiro, line 31: Any particular reason for this tabIndex?

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,73 @@
...
+      });
+      pages[show].classList.add("visible");
+    };
+  
+    togglePageVisibility();
+  
+    const togglePage = (e, node) => {
+      togglePageVisibility(node.getAttribute("data-index"));
+      e.currentTarget
+        .querySelectorAll(".nav-bar__list_item")
```
zonzujiro, line 52: First of all, you can use `querySelector('.nav-bar__list_item .active')`. Cause you should have only one active item.

Why not save `nav-bar` into a variable and then just go thru it? Something like `navList.querySelector('.active')` looks more readable and understandable than `e.currentTarget` etc.

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,73 @@
...
+    };
+  
+    navList.addEventListener("click", (e) => {
+      if (e.target.getAttribute("data-index")) {
+        togglePage(e, e.target);
+      } else if (
+        e.target.closest(".nav-bar__list_item").getAttribute("data-index")
+      ) {
+        togglePage(e, e.target.closest(".nav-bar__list_item"));
+      }
```
zonzujiro, line 64: Can be replaced only with `e.target.closest` logic. `closest` can return current item, AFAIU 🤔 Need to be checked tho.

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+  
+    document.addEventListener("keyup", (e) => {
+      if (e.code === "Space") {
+        togglePage(e, e.target);
+      }
+    });
```
zonzujiro, line 71: `Space` on `document`? Please no :) It's way too much. Space used for scrolling pages, as an example. No need in changing existing patterns without reasons.

zonzujiro (changes requested):

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: 0 this is my page overview
1 is the about page

Sadoharu: 3 - Achievement
4 - Projects
5 - Reviews

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: for control from the keyboard without a mouse. I can not use the tabindex, but i will have to add wrapper-link

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: I can remove this line of code by adding buttons

Sadoharu:
60 terms left. Maybe less?)

submissions/sadoharu/task_js_dom_/scripts.js
zonzujiro: And how buttons will help you?

Sadoharu: I can select any item from the menu using only the 'tab' and 'space' keys, without mouse

zonzujiro: Considering the fact, that it's a menu, it should be something like this:

```
nav
  ul
    li
      a
    li
      a
    li
      a
  ```
  
  In such a case with tab you will go thru anchors

Comment (Sadoharu) [reply submissions/sadoharu/task_js_dom_/scripts.js]: Instead of the a tag, I used button

File: submissions/sadoharu/task_js_dom_/scripts.js [hunk 10]
```diff
@@ -0,0 +1,61 @@
...
+        <span class="${item.icon}">
+        </span>
+        ${item.title}
+      </button>
+      <button class="nav-bar__list_button mobile ${item.icon}"></button>
+    `;
+    navList.append(liItem);
+  });
+
+  navList.querySelector(".nav-bar__list_button").classList.add("active");
```
zonzujiro, line 43: Why not do this during creation of the list? 🤔

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const togglePageVisibility = (show = 0) => {
+    pages.forEach((item) => {
+      item.classList.remove("visible");
+    });
```
zonzujiro, line 48:
```
navList.querySelector('.active').classList.remove('.active');
pages[show].classList.add("visible");
```

As an option. Feel free to ignore

submissions/sadoharu/task_js_dom_/scripts.js
zonzujiro: So what you're doing is called [magic numbers](https://codeburst.io/software-anti-patterns-magic-numbers-7bc484f40544). Please, create constant and then use it.

zonzujiro (changes requested):
Final push

Sadoharu:
> Please, create constant and then use it.
i have "pages"

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: i use forEach to create a list, and I only need one "active"

zonzujiro: So? What stops you from adding this property before adding your elements into DOM? 🤔

Sadoharu: I have them dynamically created 'li', how can i assign a class when i don't have 'li' created yet

submissions/sadoharu/task_js_dom_/scripts.js
Sadoharu: no, its the html collection index

submissions/sadoharu/task_js_dom_/scripts.js
zonzujiro: First of all, you controlling creation of them - so you have acces to strings in `forEach`. ;)
Second, you can create first item out of loop and then create others using forEach.
Third, you can create them, but not add to the DOM. Add your class and add elements to DOM after.

First option is better, as for me.

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,61 @@
...
+        ${item.title}
+      </button>
+      <button class="nav-bar__list_button mobile ${item.icon}"></button>
+    `;
+    navList.append(liItem);
+  });
+
+  navList.querySelector(".nav-bar__list_button").classList.add("active");
+
+  const togglePageVisibility = (show = 0) => {
```
artem-trubin, line 45: What Ivan meant is that this is unreadably and not intuitive. `show` doesn't tell anything about what kind of data is there and these numbers are basically random from this function perspective. You can call this variable something like `currentPage` to add more sense to it. There is also really no reason to make this function callable without passing an argument. 

About the `0` and other indexes. You're basically gathering an array of HTML elements that are already in your HTML file. But from what I see, you already figured out how to dynamically create elements from JS with your menu buttons. Why don't you do the same with pages?

submissions/sadoharu/task_js_dom_/scripts.js
```diff
@@ -0,0 +1,61 @@
...
+    liItem.setAttribute("data-index", i);
+    liItem.innerHTML = `
+      <button class="nav-bar__list_button desktop">
+        <span class="${item.icon}">
+        </span>
+        ${item.title}
+      </button>
+      <button class="nav-bar__list_button mobile ${item.icon}"></button>
+    `;
+    navList.append(liItem);
```
artem-trubin, line 40: You're reaching out to dom for every button to add it to the list. Find a way to make a single action to add all buttons at once. [hint](https://developer.mozilla.org/ru/docs/Web/API/Document/createDocumentFragment)

artem-trubin (commented):
Hey, almost there, let's fix some things.

PR #50, opened 2022-08-02
frogger

AntonKottans:
### frogger
[Demo](https://antonkottans.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/AntonKottans/frontend-nanodegree-arcade-game/tree/changes)

Please, review.


submissions/antonkottans/frogger/app.js
```diff
@@ -0,0 +1,80 @@
+const Enemy = function (x, y, speed) {
```
A-Ostrovnyy, line 1: Use JS class syntax instead of function-constructor.

submissions/antonkottans/frogger/app.js
```diff
@@ -0,0 +1,80 @@
...
+    this.y = y;
+    this.speed = speed;
+    this.sprite = "images/enemy-bug.png";
+};
+const field = {
+    x: 4,
+    y: 4,
+};
+
+Enemy.prototype.update = function (dt) {
```
A-Ostrovnyy, line 12: Avoid writing OOP code using `.prototype` use JS class for it.

submissions/antonkottans/frogger/app.js
```diff
@@ -0,0 +1,80 @@
...
+        player.x = player.start.x;
+        player.y = player.start.y;
+    }
+};
+
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const player = {
```
A-Ostrovnyy, line 33: Why do you write Enemy entity in OOP style and player not in OOP style? Try to keep the same code style throughout the project.

submissions/antonkottans/frogger/app.js
```diff
@@ -0,0 +1,80 @@
...
+            ? this.x--
+            : {};
+        if (this.y < 0) this.toStart();
+    },
+};
+
+const allEnemies = (() => {
+    const enemyRows = [60, 143, 225];
+    return [0, 1, 2].map((_, i) => {
+        return new Enemy(0, enemyRows[i], Math.random() * (200 - 60) + 60);
```
A-Ostrovnyy, line 67: `enemyRows[i]` is the same as 
```JS
[60, 143, 225].map((item) => item)
```
Also, what are the magic numbers in the array? Can you remember what those magic numbers are when you come back to this code in a year or two?

submissions/antonkottans/frogger/app.js
```diff
@@ -0,0 +1,80 @@
...
+    if (this.x >= (field.x+1) * 100) this.x = -100;
+    else this.x += dt * this.speed;
+
+    this.checkCollision();
+};
+
+Enemy.prototype.checkCollision = function () {
+    if (
+        Math.abs(player.x * 100 - this.x) <= 70 &&
+        Math.abs(player.y * 80 + 50 - this.y) < 20
```
A-Ostrovnyy, line 22: What are these magic numbers: 100, 70, 50, etc?

A-Ostrovnyy (changes requested):
Good work!
See my comments below.

AntonKottans:
all fixed

submissions/antonkottans/frogger/app.js
```diff
@@ -1,71 +1,93 @@
...
+class Enemy {
+    constructor({ x, y, speed, row }) {
+        this.x = x;
+        this.y = y;
+        this.speed = speed;
+        this.sprite = "images/enemy-bug.png";
+        this.row = row;
+    }
+    checkCollision() {
+        const distanceOfCollision = 70;
```
A-Ostrovnyy, line 21: How you get this value?)

A-Ostrovnyy (changes requested):
Awesome!

submissions/antonkottans/frogger/app.js
AntonKottans: its difference in x-coord  of bug and player while visual connection

A-Ostrovnyy: I think this approach is not flexible, it is better to check collisions with respect to coordinates not only on the x-axis and on the y-axis as well.

A-Ostrovnyy (changes requested):
One small issue

submissions/antonkottans/frogger/app.js
```diff
@@ -1,37 +1,35 @@
...
         this.x = x;
         this.y = y;
         this.speed = speed;
         this.sprite = "images/enemy-bug.png";
-        this.row = row;
     }
     checkCollision() {
-        const distanceOfCollision = 70;
-        const sameRow = () => player.y === this.row;
+        const distanceOfXCollision = field.squareWidth / 1.3;
```
A-Ostrovnyy, line 20: Magic numbers are [bad](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

submissions/antonkottans/frogger/app.js
```diff
@@ -1,37 +1,35 @@
...
+        const distanceOfYColision = 1;
+        const sameRow = () => Math.abs(player.y - this.y) < distanceOfYColision;
```
A-Ostrovnyy, line 22: Try to calculate collision using player and enemy coordinates without `sameRow` method.

A-Ostrovnyy (changes requested):
Sorry for the late response. lost your PR.

A-Ostrovnyy (approved):
Awesome!

PR #92, opened 2022-08-08
Frogger js core code

AsaMitaka:
### Object oriented JS

[Demo](https://asamitaka.github.io/frontend-nanodegree-arcade-game)
[Code](https://github.com/AsaMitaka/frontend-nanodegree-arcade-game/blob/master/js/app.js)

Please, review.

### DIFF
--- a/submissions/asaMitaka/frogger/app.js
+++ b/submissions/asaMitaka/frogger/app.js
@@ -1,6 +1,6 @@
-const newP = document.createElement('p');
-newP.textContent = `Your score: 0 `;
-document.body.append(newP);
+const score = document.createElement('p');
+score.textContent = `Your score: 0 `;
+document.body.append(score);
 
 const enemyStat = {
     width: 98,
@@ -76,8 +76,9 @@ let allEnemies = allEnemyVal.map(
 const Player = function(posX, posY) {
     this.posX = posX;
     this.posY = posY;
+    this.stepX = 100;
+    this.stepY = 90;
     this.stat = 0;
-    this.step = 100;
 
     this.width = 80;
     this.height = 80;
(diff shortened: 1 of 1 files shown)


submissions/asaMitaka/frogger/app.js
```diff
@@ -0,0 +1,169 @@
+// Enemies our player must avoid
```
artem-trubin, line 1: Seems like an irrelevant comment here.

submissions/asaMitaka/frogger/app.js
```diff
@@ -0,0 +1,169 @@
...
+var newP = document.createElement('p');
```
artem-trubin, line 2: You should avoid using `var`s. Most of the time to store a value `const` is enough and only if you are 100% sure that you'll change this value in code later use `let`.

submissions/asaMitaka/frogger/app.js
```diff
@@ -0,0 +1,169 @@
...
+newP.textContent = `Your score: 0 `;
+document.body.append(newP);
+
+var enemyStat = {
```
artem-trubin, line 6: There is no reason for a separate object that stores information about enemies since you already has a whole class for them. Store these values there.

submissions/asaMitaka/frogger/app.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+Player.prototype.update = function() {
+    this.win();
+}
+
+Player.prototype.win = function() {
+    if (field.min >= this.posY) {
+        console.log(field.min, this.posY);
```
artem-trubin, line 111: `console.log` is fine for debugging but should be removed from the final version of the code.

artem-trubin (commented):
Hey! There are some improvements that can be done, but in general, nice work! 

You've chosen to make the character go not cell-by-cell but with smaller steps which are fine but it's a little strange that it can go so low, almost off the level, and stops going left so early. Please work on the limits of where the character can go.

Also, check comments below.

artem-trubin:
Did you forget to update your own repo? Github Pages demo still has problems listed in my previous review.

OleksiyRudenko (changes requested):
Also follow the requirements here https://github.com/kottans/frontend-2022-homeworks/pull/92#issuecomment-1216478616

Feel free asking questions here or in Students' chat if anything is not really clear.

When done, please confirm explicitly that all requirements are met.

artem-trubin:
Great job!

PR #93, opened 2022-08-08
OOP exercise

AsaMitaka:
### OOP exercise

[Demo](https://asamitaka.github.io/a-tiny-JS-world/index2.html) |
[Code](https://github.com/AsaMitaka/a-tiny-JS-world/blob/master/indexClass.js)

Please, review.


submissions/asaMitaka/OOP exercise/indexClass.js
```diff
@@ -0,0 +1,96 @@
...
+        this.hands = 0;
+        this.saying = saying;
+    }
+
+    aboutMe() {
+        return `${this.saying} My name is ${this.name}. I'm ${this.species} ${this.gender}. I have ${this.legs} legs.`;
+    }
+}
+
+let john = new Man('John', 'Hello');
```
artem-trubin, line 87: These are consts.

artem-trubin (commented):
Hey! Let's improve your work, check comments below.

submissions/asaMitaka/OOP exercise/indexClass.js
```diff
@@ -0,0 +1,70 @@
...
+        this.gender = 'male';
+        this.saying = saying;
+    }
+}
+
+class Woman extends Human {
+    constructor (name, saying = 'Hi') {
+        super(name, saying);
+        this.gender = 'female';
+        this.saying = saying;
```
artem-trubin, line 41: There are a lot of repeating actions, for example - `saying`. When you create `Woman`, you pass a string as a `saying` to it. It gets it as a constructor parameter and then passes it to the `super` constructor. Later, on the `41` line it assigns this value to the `this. saying` property. But what does `super` do with it? Well, here, basically there's broken logic since the second parameter of the `Human` constructor is gender. So you pass saying as gender. And you also rewrite `gender` on line 40. Do you see what I'm saying? Please restructure your code, maybe even from scratch to make sure that every class is responsible for its exclusive properties and if something needs to be set from outside (like names or genders) it is passed properly.

artem-trubin (commented):
Hi! It is an improvement for sure, but there are still problems. Please read the comment below.

artem-trubin (commented):
Something went wrong since the last review. Where did the `Inhabitant` class go? It looks like you redid your structure, but there are some flaws now. Please rewrite it again, considering these points:
- Your whole hierarchy should have **one** common ancestor.
- If you see a property present among all descendants of some ancestor - it should be lifted up the hierarchy tree and assigned to the ancestor.
- Do not rewrite the method for getting information *entirely*. If your ancestor tells about some property and you want to add something to it - add it, don't rewrite the whole message.

artem-trubin:
That's it, well done!

PR #94, opened 2022-08-08
Memory pair game

AsaMitaka:
### [Memory pair game](https://kottans.org/frontend/tasks/memory-pair-game.html)

[Demo](https://asamitaka.github.io/kottans-frontend/memory-pair-game) |
[Code](https://github.com/AsaMitaka/kottans-frontend/tree/main/memory-pair-game)

Please, review.


submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+let previousClicked = undefined;
+let currentClicked = undefined;
+let flipped = 0;
+let lockBoard = false;
+
+// Clicked item change img
+function clickedItem(event) {
+    if (this === previousClicked) return;
```
zonzujiro, line 64: Please avoid such things in 2022. :) While it's a completely legit and a working option, it's opaque while reading the code. The reader will need to go and find use cases of your function to understand what values can have `this`.

Instead, you can just use `event.target`.

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,106 @@
...
+    if (lockBoard) return;
+    if (!event.target.classList.contains('cardItem__front')) return; 
+
+    event.target.parentElement.classList.add('flip');
```
zonzujiro, line 68: Don't use constructions like `children[0]`, `firstElementChild`, `parentNode`, `nextSibling`, etc. In such a way, you rely on the order of DOM elements. So in case when you will have changed the design - your code will be broken. Which is bad, obviously. Use `querySelector` or `closest`, if in event, instead.

submissions/asaMitaka/Object oriented JS/app.js
```diff
@@ -0,0 +1,169 @@
+// Enemies our player must avoid
```
zonzujiro, line 1: Looks like a redundant file here. You should delete it.

zonzujiro (changes requested):

zonzujiro:
Ah, and use prettier please :)

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
+let content = document.querySelector('.content');
+const backImg = 'https://i.pinimg.com/originals/13/25/05/132505ba3238e79c00034c905b2ca045.jpg';
+
+let arrOfCards = [
```
zonzujiro, line 4: Common mistake - variable name. Please read review doc :)

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
...
+    {
+        img: 'https://static3.depositphotos.com/1000695/119/i/450/depositphotos_1190337-stock-photo-figure-eight.jpg',
+        id: 8,
+    },
+];
+
+let shuffledArr = shuffleArr([...arrOfCards, ...arrOfCards]);
+
+shuffledArr.forEach((el) => {
+    content.innerHTML += renderItem(el);
```
zonzujiro, line 42: Common mistake - adding elements in the loop.

https://kottans.org/documentation/docs/doc/code-review/ please check the doc :)

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
...
+            <img class='cardItem__front' src='${backImg}'/>
+            <img class='cardItem__back' src='${item.img}'/>
+        </div>
+    `;
+}
+
+let previousClicked = undefined;
+let currentClicked = undefined;
+let flipped = 0;
+let lockBoard = false;
```
zonzujiro, line 69: Codestyle - boolean variables should be called with `is`, `has`, etc. In your case it should be `isBoardLocked`.

And `lockBoard` is verb. It's not a function, so it should be noun

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
...
+
+function clickedItem(event) {
+    if (event.target === previousClicked) return;
+    if (lockBoard) return;
+    if (!event.target.classList.contains('cardItem__front')) return;
```
zonzujiro, line 74: You can make one if for this.

And I would move some checks out of function. No need to call listener in general if board was locked.

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
...
+
+function clickedItem(event) {
+    if (event.target === previousClicked) return;
+    if (lockBoard) return;
+    if (!event.target.classList.contains('cardItem__front')) return; 
+
+    let parentElement = event.target.closest('.cardItem');
+    parentElement.classList.add('flip');
+
+    if (previousClicked === undefined) {
```
zonzujiro, line 79: Just `!previousClicked`. It's shorter and will fit your needs

submissions/asaMitaka/Memory pair game/script.js
zonzujiro: And function names should be verbs

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -0,0 +1,113 @@
...
+        let j = Math.floor(Math.random() * (i + 1));
+        [array[i], array[j]] = [array[j], array[i]];
+    }
+
+    return array;
+}
+
+function renderItem(item) {
+    return `
+        <div class='cardItem' data-attribute='${item.id}'>
```
zonzujiro, line 59: `data-attribute` to generic. Please use something more concrete

zonzujiro (changes requested):

submissions/asaMitaka/Memory pair game/script.js
```diff
@@ -36,14 +40,10 @@ let arrOfCards = [
     },
 ];
 
-let shuffledArr = shuffleArr([...arrOfCards, ...arrOfCards]);
-
-shuffledArr.forEach((el) => {
-    content.innerHTML += renderItem(el);
-});
+let shuffledCardList = shuffleArr([...cardsList, ...cardsList]).map((el) => content.innerHTML += renderItem(el));
```
zonzujiro, line 43: Still same mistake - you adding elements inside loop. Don't do that :)

zonzujiro (changes requested):
One last thing. Also, please resolve outdated conversations

zonzujiro (approved):

PR #95, opened 2022-08-08
Friends app

AsaMitaka:
### Friends App

[Demo](https://asamitaka.github.io/kottans-frontend/friends-app) |
[Code](https://github.com/AsaMitaka/kottans-frontend/tree/main/friends-app)

Please, review.


submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
+const url = 'https://randomuser.me/api/?results=12';
+
+let aToZLast = document.querySelector('.sortedAtoZLast');
```
artem-trubin, line 3: You almost exclusively use `let` variables while most of them should be `const`. Please review your code and change them.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+let zToALast = document.querySelector('.sortedZtoALast');
```
artem-trubin, line 4: There are a lot of `querySelector`s here. Instead, you can just get your `form` element and then access any field of the form at any time by its `name`. 

So for example, you can have a couple of radiobuttons that share the same `name`, let's say "gender". Each of them should have a `value` attribute, like `male` and `female` for example.

So then you can just go
```js
const form = document.querySelector("#myForm");
console.log(form.gender) // this should return male\female based on checked radiobutton
```

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+    renderAllItemsToPage(resultArr);
+}
+
+
+function saveUsers(arr, resArr) {
+    for (let i = 0; i < arr.length; i++) {
+        resArr.push(arr[i]);
+    }
+    basicArr1 = resultArr.slice();
+    // basicArr2 = JSON.parse(JSON.stringify(resultArr));
```
artem-trubin, line 34: Remove commented code from the final version.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+let content = document.querySelector('.articleBlock');
+let reset = document.querySelector('.reset');
+let male = document.querySelector('#male');
+let woman = document.querySelector('#female');
+let all = document.querySelector('#all');
+let search = document.querySelector('#searchBar');
+let minNumber = document.querySelector('#minNumber');
+let maxNumber = document.querySelector('#maxNumber');
+
+let basicArr1 = [];
```
artem-trubin, line 16: `basicArr1` doesn't seem like a correct name. You should avoid using data type in the name of the variables, it indicates that the name is not descriptive enough. Like `resultArr` could just be `friends` or `users`. Plural name assumes that this is an array by itself.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+    let response = await fetch(url);
+    let friendsArr = await response.json();
+
+    saveUsers(friendsArr.results, resultArr);
+
+    renderAllItemsToPage(resultArr);
+}
+
+
+function saveUsers(arr, resArr) {
```
artem-trubin, line 29: What does this function do, exactly?
It seems like it just copies the whole `arr` to `resArr` and then does the same for `resultArr` and `basicArr1`. It's not readable and seems like it doesn't make any sense at all. You can just save users right where you get your response from API.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+let search = document.querySelector('#searchBar');
+let minNumber = document.querySelector('#minNumber');
+let maxNumber = document.querySelector('#maxNumber');
+
+let basicArr1 = [];
+let resultArr = [];
+
+async function getResponse(url) {
+    let response = await fetch(url);
+    let friendsArr = await response.json();
```
artem-trubin, line 21: It's not even ar Arr, it's an object that has an array in the `results` field.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+
+function saveUsers(arr, resArr) {
+    for (let i = 0; i < arr.length; i++) {
+        resArr.push(arr[i]);
+    }
+    basicArr1 = resultArr.slice();
+    // basicArr2 = JSON.parse(JSON.stringify(resultArr));
+}
+
+function renderItem(item) {
```
artem-trubin, line 37: Misleading name, this function doesn't `render` anything, it returns an HTML code for an element.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,177 @@
...
+            </p>
+            <p>Age: ${item.dob.age}</p>
+            <a href='${item.phone}'>${item.phone}</a>
+            <p class='content__item-country'>${item.location.country}, ${item.location.city}</p>
+            <div class='content__item-gender'>${item.gender}</div>
+        </div>
+    `;
+}
+
+aToZLast.addEventListener('click', sortedAtoZLast);
```
artem-trubin, line 53: You should avoid using a lot of event listeners. This task can be done with just one that is placed on the whole form.

artem-trubin (commented):
Hi! Great work for the first iteration. Let's fix some code.

For starters, you have a bug. Your sorts and filters don't persist. If you sort by name and then filter, sorting disappears. If you filter by a certain name and then filter by gender, it shows all women/men instead of only those that have the filtered name. Your right side of the app should always represent the results of the whole form state, not only the last clicked filter/sorting.

Secondly, from the UX perspective, your sorting should be displayed as radio buttons and it should be visible what sorting is selected at the current time.

Also, please check comments below.

artem-trubin:
The bug still persists. Write something in the search field, and make sure a user or several of them are found by the name. Then, click different sorts and filters. The user list starts to ignore what's written in the search field. And it shouldn't. As I said, at any point in time, the list of users on the right should represent the state of the form on the left.

artem-trubin (commented):
It seems that GitHub is still pending my review, so I'm going to post it here as a review. I saw your changes, but there are still bugs. Please make a lot of testing with your form. Sometimes First Name sorting just stops working, when you change age it doesn't really filter anything. And why is the label for age limits say `Age 10 - 100`? You can change it for whatever, not only 10\100? :)
Just some advice, maybe change your name filtration to filter through both first and last names? Have you seen anywhere separate filtrations for name and surname? You can just stitch them to one string and filter them like this.

submissions/asaMitaka/FriendsApp/index.html
```diff
@@ -0,0 +1,71 @@
...
+                <form class="asideBlock__items" name="last">
+                    <p>By Last Name</p>
+                    <input type="radio" name="last" id="lastAToZ" value="lastAToZ" class="last">
+                    <label for="lastAToZ">A-Z</label>
+                    <input type="radio" name="last" id="lastZToA" value="lastZToA" class="last">
+                    <label for="lastZToA">Z-A</label>
+                </form>
+            
+                <form class="asideBlock__items" name="minMax">
+                    <p>Age 10 - 100</p>
```
artem-trubin, line 35: Why is it exactly `10 - 100`?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    try {
+        response = await fetch(url);
+        if (!response.ok) {
+            throw new Error(response.statusText);
+        } 
+    } catch (err) {
+        console.log(err);
+        resetPage();
+    }
+    let jsonData = await response.json();
```
artem-trubin, line 20: const?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
+const url = 'https://randomuser.me/api/?results=12';
+const form = document.forms;
+const content = document.querySelector('.articleBlock');
+
+let friendsCopy = [];
+let friends = [];
+let isSearch = false;
+
+async function getResponse(url) {
+    let response;
```
artem-trubin, line 10: You don't need an outside variable for `response`. Do the `.json()` thing right in the `try` block. You have a faulty logic here, you're getting data from a response outside of `try` in any scenario, even if the data isn't there because you ignore the `catch` block. The only reason your program doesn't crash is that if there _is_ an error you just reset your page, but let's not rely on that. If there is no data - don't extract it from the response.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    }
+    let jsonData = await response.json();
+
+    friendsCopy = jsonData.results;
+    friends = jsonData.results;
+
+    renderAllItemsToPage(friends);
+}
+
+function creatingDivItem(item) {
```
artem-trubin, line 28: What `div` item? What is the content of this item? Use more descriptive names, please.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+            </p>
+            <p>Age: ${item.dob.age}</p>
+            <a href='${item.phone}'>${item.phone}</a>
+            <p class='content__item-country'>${item.location.country}, ${item.location.city}</p>
+            <div class='content__item-gender'>${item.gender}</div>
+        </div>
+    `;
+}
+
+form.last.addEventListener('click', function(e) {
```
artem-trubin, line 44: Don't rely on the fact that your form is the last one on the page. Reason: there are a lot of browser extensions that modify the dom, for example, Grammarly adds its script to the end of DOM to work. That is why we don't style by tags and don't do JS by tags - because you never know what could go wrong with the DOM structure. Give your form an ID and do a `querlySelector`.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    if (e.target && e.target.classList.contains('last')) {
+        checkLastSort(e);
```
artem-trubin, line 46: You only need `target` from event object, you can destructure it on the start of the function like
```js
function({ target }) { ...
```

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+        case "lastAToZ":
+            sortedAtoZLast();
+            break;
+        case "lastZToA":
+            sortedZtoALast();
+            break;
+    }
+}
+
+form.reset.reset.addEventListener('click', resetPage);
```
artem-trubin, line 62: Consider resetting the form state and rerendering users instead of reloading the whole page. You can also use [this](https://developer.mozilla.org/ru/docs/Web/API/HTMLFormElement/reset).

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+
+form.age.addEventListener('click', function(e) {
```
artem-trubin, line 64: You add a lot of event listeners for separate items of one form. Please combine them all into one event listener for the whole form. If you use the correct event type, this event listener will fire each time you press a button in your search input and each time a radiobutton is pressed.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    if (e.target && e.target.classList.contains('age')) {
+        sortByAge(e);
+    }
+});
+
+function sortByAge(event) {
```
artem-trubin, line 70: Same thing with `target`.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+        case "ageMore":
+            sortedAgeMore();
+            break;
+        case "ageLess":
+            sortedAgeLess();
+            break;
+    }
+}
+
+function sortedAgeMore() {
```
artem-trubin, line 82: Why `sorted`? It's not a verb. You had `sort` in your other function.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+            break;
+        case 'all':
+            renderAll();
+            break;
+    }
+}
+
+function renderMale() {
+    friends = friendsCopy.slice();
+    friends = friends.filter(item=> item.gender === 'male');
```
artem-trubin, line 124: We usually place a space around `=>` in arrow functions. So `item => ...`

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    renderAllItemsToPage(friends);
+}
+
+function renderFemale() {
+    friends = friendsCopy.slice();
```
artem-trubin, line 129: `.slice()` without arguments just returns the same array, why is it here?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+        case 'female':
+            renderFemale();
+            break;
+        case 'all':
+            renderAll();
+            break;
+    }
+}
+
+function renderMale() {
```
artem-trubin, line 122: Your `renderMale`, `renderFemale` and `renderAll` functions are very similar. Please combine them.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    
+    if (maxNumber.value > 100) {
+        maxNumber.value = 100;
+        return;
+    } 
+
+    findByAge();
+}
+
+function findByAge() {
```
artem-trubin, line 162: In terms of JS, `find` means 1 result and `filter` means multiple results.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+        }
+    });
+
+    renderAllItemsToPage(friends);
+}
+
+form.search.search.addEventListener('input', searchFunc);
+
+function searchFunc() {
+    isSearch = true;
```
artem-trubin, line 193: You don't need a variable to track if there is something in the input. Call this function every time something changes, just don't do anything here if the input is empty.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+    let val = form.search.search.value.toLowerCase();
+    if (val && val.length > 0) {
+        content.innerHTML = '';
```
artem-trubin, line 196: Why is this here? It's a searching function, it should return an array of users that match search input, not do things in the DOM.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+function renderAllItemsToPage(arr) {
+    if (form.search.search.value.length === 0) {
+        isSearch = false;
+    }
+
+    content.innerHTML = '';
+
+    let acc = '';
```
artem-trubin, line 226: Instead of creating a string of elements, it would be better to use the `map` method and create an array of elements, just join them before rendering. This would allow you to manipulate this array in the future if needed.

artem-trubin (changes requested):
Hey, nice work. Let's improve it.

You have some trouble with setting clear responsibilities for your functions. If you have a function that says `filterUsersByName`, it should receive an array to filter and what to search for as parameters and return a filtered array in the end. And that's it. You then do anything you want with this array - sort it, render it, whatever, but this `filterUsersByName` function should not care, it has done its work already. You can make a separate function for rendering and just pass data to it. You have a lot of functions that do some action, like sorting **AND** they also render to DOM. Such approach invites mistakes and reduces code readability.

submissions/asaMitaka/FriendsApp/index.html
```diff
@@ -0,0 +1,63 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Friends App</title>
+    <link rel="stylesheet" href="style.css">
+</head>
+<body>
+    <div class="container">
+        <div class="header">
```
artem-trubin, line 12: Why did you go away from semantic `header` and `footer` tags?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,87 @@
+const url = 'https://randomuser.me/api/?results=12';
+const content = document.querySelector('.articleBlock');
+const form = document.forms;
+let friends = [];
+let friendsCopy = [];
+
+const data = (url) => fetch(url)
+    .then(handleErrors)
+    .then(res => res.json())
+    .then(res => res.results)
```
artem-trubin, line 10: Getting a property from an object is not an asynchronous task, you don't need another `.then` after this one.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,87 @@
+const url = 'https://randomuser.me/api/?results=12';
+const content = document.querySelector('.articleBlock');
+const form = document.forms;
```
artem-trubin, line 3: Why do you have 2 forms? You should use 1 form and get it by id or something, don't rely on `document.forms`.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,87 @@
...
+    return res;
+}
+
+document.querySelector('.asideBlock').addEventListener('click', sortBy);
+function sortBy({target}) {
+    const id = target.id;
+
+    switch(id) {
+        case 'ageMore':
+            friendsCopy = friendsCopy.sort((a, b) => a.dob.age - b.dob.age);
```
artem-trubin, line 30: Look at the functions that you use for sorting. They are identical, except you change parameter order. Maybe abstract it to not repeat yourself.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,87 @@
...
+        case 'ageLess':
+            friendsCopy = friendsCopy.sort((a, b) => b.dob.age - a.dob.age);
+            break;
+        case 'lastAToZ':
+            friendsCopy = friendsCopy.sort((a, b) => a.name.last !== b.name.last ? a.name.last < b.name.last ? -1 : 1 : 0);
+            break;
+        case 'lastZToA':
+            friendsCopy = friendsCopy.sort((a, b) => a.name.last !== b.name.last ? a.name.last > b.name.last ? -1 : 1 : 0);
+            break;
+        case 'male':
```
artem-trubin, line 41: This is filtering, it should not belong to the sorting function.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,87 @@
...
+
+    let acc = '';
+    arr.forEach(item => {
+        acc += creatingProfileCard(item);
+    });
+
+    content.innerHTML = acc;
+}
+
+data(url);
```
artem-trubin, line 87: Add empty lines at the end of all files.

artem-trubin (changes requested):
You have a bug now, if you select any gender, even the one that is already selected your sorting drops. Also, what happened to the name search and age filtration?

submissions/asaMitaka/FriendsApp/index.html
```diff
@@ -0,0 +1,71 @@
...
+                    <div class="asideBlock__items">
+                        <p>By Last Name</p>
+                        <input type="radio" name="sort" id="lastAToZ">
+                        <label for="lastAToZ">A-Z</label>
+                        <input type="radio" name="sort" id="lastZToA">
+                        <label for="lastZToA">Z-A</label>
+                    </div>
+                </form>
+    
+                <form name="gender" id="genderForm">
```
artem-trubin, line 38: You're still using separate forms. It makes your code more complicated in the JS. Why not tie it up in one form?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,161 @@
...
+    gender: null,
+    sort: null,
+    reset() {
+        this.search = null;
+        this.gender = null;
+        this.sort = null;
+    }
+}
+
+function filterBy(id, usersToFilter) {
```
artem-trubin, line 35: I believe the only id you have there is somehow connected with gender, so `filterBy` is actually `filterByGender`.

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,161 @@
...
+    switch(id) {
```
artem-trubin, line 36: You can shorten this switch like this:

```js
switch(id) {
  case 'male':
  case 'female':
    usersToFilter = userstoFilter.filter(item => item.gender === id);
    break;
  case 'all':
  ...
}
```

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,161 @@
...
+        case 'male':
+            usersToFilter = usersToFilter.filter(item => item.gender === 'male');
+            break;
+        case 'female':
+            usersToFilter = usersToFilter.filter(item => item.gender === 'female');
+            break;
+        case 'all':   
+            usersToFilter = [...friends];
```
artem-trubin, line 44: First of all, it would be more readable to create a new variable, something like `filteredUsers` in the beginning of this function to store the returned result. Secondly, if you're getting this from `friends` what's the point of passing some array in the function argument?

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,161 @@
...
+            break;
+    }
+    return usersToFilter;
+}
+
+function sortAge(a, b) {
+    return a.dob.age - b.dob.age;
+}
+
+function sortAlph(a, b) {
```
artem-trubin, line 54: `sortByName`? And `sortByAge`

submissions/asaMitaka/FriendsApp/script.js
```diff
@@ -0,0 +1,161 @@
+const url = 'https://randomuser.me/api/?results=12';
+const content = document.querySelector('.articleBlock');
+const ageForm = document.querySelector('#ageForm');
+const genderForm = document.querySelector('#genderForm');
+const search = document.querySelector('#search');
+let friends = [];
+let friendsCopy = [];
+
+const data = (url) => fetch(url)
```
artem-trubin, line 9: `getData`?

artem-trubin (changes requested):
Please check comments below.

artem-trubin:
Great work!

PR #96, opened 2022-08-08
create task js-dom

semenjob:
### Document Object Model

[Demo](https://semenjob.github.io/JS-DOM/) |
[Code base](https://github.com/semenjob/JS-DOM)

Please, review.


submissions/semenjob/script/index.js
```diff
@@ -0,0 +1,52 @@
+window.addEventListener('DOMContentLoaded', () => {
+  initMusicApp()
+})
+
+function initMusicApp() {
+  const tabs = document.querySelectorAll('.wrapper__item'), contentTitle = document.querySelector('.music__title'), contentDescr = document.querySelector('.music__descr'), parentItems = document.querySelector('.wrapper__list')
+
```
madmaxWMFU, line 7: The load function should only have initial function calls, description of some variables, possibly initial checks

submissions/semenjob/index.html
```diff
@@ -0,0 +1,47 @@
...
+  <body>
+    <div class="bg">
+      <div class="container">
+        <div class="container_title">Music</div>
+        <div class="wrapper">
+          <div class="wrapper__list">
+            <div class="wrapper__item">Pop</div>
+            <div class="wrapper__item">K-Pop</div>
+            <div class="wrapper__item">Rock</div>
+            <div class="wrapper__item">Alternative Rock</div>
```
madmaxWMFU, line 23: Show more your skills in creating and drawing DOM elements.

madmaxWMFU (changes requested):
Beautiful work. Only you have not fulfilled all the requirements of the assignment (creating elements).

submissions/semenjob/script/index.js
zonzujiro, line 7: Please, use Prettier :) https://prettier.io/playground/

It will make code easier to read

submissions/semenjob/script/index.js
```diff
@@ -0,0 +1,52 @@
...
+  parentItems.addEventListener('click', (event) => {
+      const target = event.target
+
+    if (target && target.classList.contains('wrapper__item')) {
+        tabs.forEach((item, i) => {
+      if (target == item) {
+          contentTitle.innerText = musicStyle[i].title
+          contentDescr.innerText = musicStyle[i].desc
+        }
```
zonzujiro, line 47: Also prettier. Wrong intendation

submissions/semenjob/script/index.js
```diff
@@ -0,0 +1,52 @@
...
+        title: 'Hip-hop and Rap',
+        desc: "Hip hop, also hip hop music or rap music, is a genre of popular music developed in the United States by African Americans and Hispanics living in the poorer parts of the Bronx in New York City in the 1970s.",
+    },
+  ]
+
+  parentItems.addEventListener('click', (event) => {
+      const target = event.target
+
+    if (target && target.classList.contains('wrapper__item')) {
+        tabs.forEach((item, i) => {
```
zonzujiro, line 43: That's not how you should use DOM. You're using it to find clicked element and then replace the content inside of it.

Instead of using `NodeList#forEach`, you should understand what item was clicked and then get the necessary content from an array. Without traversing DOM elements.

In other words - separate rendering of elements from working with data of elements.

zonzujiro (changes requested):

submissions/semenjob/index.html
semenjob: ok i understand what you mean

submissions/semenjob/script/index.js
semenjob: ok

semenjob: but here, to be honest, I didn’t understand how to solve this issue, is there any kind of thing?

submissions/semenjob/script/index.js
semenjob: this is Prettier strange settings

submissions/semenjob/script/index.js
```diff
@@ -0,0 +1,55 @@
+window.addEventListener("DOMContentLoaded", () => {
+  const contentTitle = document.querySelector(".music__title"),
+    contentDescr = document.querySelector(".music__descr"),
+    parentItems = document.querySelector(".wrapper__list");
+
+  const musicStyle = [
```
zonzujiro, line 6:
```suggestion
  const musicStyles = [
```

It's an array :)

submissions/semenjob/script/index.js
```diff
@@ -0,0 +1,55 @@
...
+    return `<div class="wrapper__item">${title}</div>`;
+  };
+
+  const element = musicStyle.map(({ title }) => createStyleList(title));
+  parentItems.innerHTML = element.join("");
+
+  parentItems.addEventListener("click", function (event) {
+    const target = event.target;
+    if (target.classList.contains("wrapper__item")) {
+      const targetItem = target.innerText;
```
zonzujiro, line 47: Don't rely on things like `textContent` or `innerText` - data in this fields is used only for rendering and usually can be changed easily.  As an example, if I will translate your site with google translate, this will be broken. :)

For your "dev goals" you should use some "dev thingies". As an example - `data-*` attribute.

zonzujiro (changes requested):

submissions/semenjob/script/index.js
semenjob: In my translation, when you click on the text itself, it sometimes does not want to switch. With what it can be connected?

madmaxWMFU (approved):
Try to render the content in the same way as the menu was rendering.

zonzujiro (approved):

submissions/semenjob/dom/style/style.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.container {
+  padding: 80px 0 0 0;
+  display: block;
+  width: 80%;
+  margin: 0 auto;
+}
+
+.wrapper {
```
serentinos, line 33: Я б добавив невеличкий padding, border-radius, та напівпрозорий бекграунд, щось типу такого:
`   background-color: rgb(238,130,238, 0.3);
    padding: 20px;
    border-radius: 20px;`
виглядає набагато приємніше)

PR #117, opened 2022-08-10
Anton tjsw post oop

AntonKottans:
OOP exercise

[Demo](https://antonkottans.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/AntonKottans/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/antonkottans/tjsw-post-oop/index.js
+++ b/submissions/antonkottans/tjsw-post-oop/index.js
@@ -1,176 +1,147 @@
-const KEYS_VALUE_SEPARATOR = ": ";
-const PROPS_SEPARATOR = "; ";
-
 class Creature {
-    constructor({ name, gender, friends = [], species, saying }) {
-        this.name = name;
-        this.species = species;
-        this.gender = gender;
-        this.friends = friends;
-        this.saying = saying;
-    }
+  constructor({ name, gender, friends = [], species, saying }) {
+    this.name = name;
+    this.species = species;
+    this.gender = gender;
+    this.friends = friends;
+    this.saying = saying;
+  }
 
-    createArrayOfProps() {
-        return ["name", "species", "gender", "saying", "friends"]
-            .map((key) => {
-                if (key != "friends")
-                    return key + KEYS_VALUE_SEPARATOR + this[key];
-                return this.friends.length
-                    ? "friends: " +
-                          this.friends.map((friend) => friend.name).join(", ")
-                    : "";
-            })
-            .filter((_) => _);
-    }
+  propsToString() {
+    return `Name: ${this.name}; Species:${this.species}; Gender:${
+      this.gender
+    }; Saying:${this.saying}; Friends:${this.friendsNameToString()}`;
+  }
 
-    addFriends(...newFriends) {
-        if (newFriends.length > 0)
-            this.friends = [...this.friends, ...newFriends];
-    }
+  friendsNameToString() {
+    return this.friends.length === 0
+      ? "no friends"
+      : this.friends.map((friend) => friend.name).join(", ");
+  }
+
+  addFriends(...newFriends) {
+    this.friends = [...this.friends, ...newFriends];
+  }
 }
 
 class Animal extends Creature {
-    constructor({ name, gender, friends, paws, species, saying }) {
-        super({ name, gender, friends, species, saying });
-        this.paws = paws;
-    }
+  constructor({ name, gender, friends, paws, species, saying }) {
+    super({ name, gender, friends, species, saying });
+    this.paws = paws;
+  }
 
-    createArrayOfProps() {
-        return [
-            ...super.createArrayOfProps(),
-            "paws" + KEYS_VALUE_SEPARATOR + this["paws"],
-        ];
-    }
+  propsToString() {
+    return `${super.propsToString()}; Paws: ${this.paws}`;
+  }
 }
 
 class Human extends Creature {
-    constructor({ name, gender, friends, legs, hands, saying }) {
-        super({ name, gender, friends, species: "human", saying });
-        this.legs = legs;
-        this.hands = hands;
-    }
+  constructor({ name, gender, friends, legs, hands, saying }) {
+    super({ name, gender, friends, species: "human", saying });
+    this.legs = legs;
+    this.hands = hands;
+  }
 
-    createArrayOfProps() {
-        return [
-            ...super.createArrayOfProps(),
-            ...["legs", "hands"].map(
-                (key) => key + KEYS_VALUE_SEPARATOR + this[key]
-            ),
-        ];
-    }
+  propsToString() {
+    return `${super.propsToString()}; Legs: ${this.legs}; Hands: ${this.hands}`;
+  }
 }
 
 class Dog extends Animal {
-    constructor({
-        name,
-        gender,
-        friends,
-        paws,
-        tail = "ordinary tail",
-        saying,
-    }) {
-        super({ name, gender, friends, paws, tail, species: "dog", saying });
-        this.tail = tail;
-    }
+  constructor({ name, gender, friends, paws, tail = "ordinary tail", saying }) {
+    super({ name, gender, friends, paws, tail, species: "dog", saying });
+    this.tail = tail;
+  }
 
-    createArrayOfProps() {
-        return [
-            ...super.createArrayOfProps(),
-            "tail" + KEYS_VALUE_SEPARATOR + this.tail,
-        ];
-    }
+  propsToString() {
+    return `${super.propsToString()}; Tail: ${this.tail}`;
+  }
 }
 
 class Cat extends Animal {
-    constructor({
-        name,
-        gender,
-        friends,
-        paws,
-        tail = "ordinary tail",
-        saying = "meow!",
-    }) {
-        super({ name, gender, friends, paws, tail, species: "cat", saying });
-        this.saying = saying;
-        this.tail = tail;
-    }
+  constructor({
+    name,
+    gender,
+    friends,
+    paws,
+    tail = "ordinary tail",
+    saying = "meow!",
+  }) {
+    super({ name, gender, friends, paws, tail, species: "cat", saying });
+    this.saying = saying;
+    this.tail = tail;
+  }
 
-    createArrayOfProps() {
-        return [
-            ...super.createArrayOfProps(),
-            "tail" + KEYS_VALUE_SEPARATOR + this.tail,
-        ];
-    }
+  propsToString() {
+    return `${super.propsToString()}; Tail: ${this.tail}`;
+  }
 }
 
 class CatWoman extends Human {
-    constructor({
-        name,
-        gender,
-        friends,
-        legs,
-        hands,
-        tail = "ordinary Cat-woman tail",
-        saying = "meow!",
-    }) {
-        super({
-            name,
-            gender,
-            friends,
-            legs,
-            hands,
-            species: "cat-woman",
-        });
-        this.tail = tail;
-        this.saying = saying;
-    }
+  constructor({
+    name,
+    gender,
+    friends,
+    legs,
+    hands,
+    tail = "ordinary Cat-woman tail",
+    saying = "meow!",
+  }) {
+    super({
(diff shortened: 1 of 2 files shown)


submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,140 @@
...
+                    if (typeof this[key] === "function")
+                        return output.concat(this[key]());
+                    else return output.concat(this[key]);
+                } else return output;
+            }, [])
+            .join(", ");
+    }
+}
+class Animal extends Creature {
+    constructor(props) {
```
OleksiyRudenko, line 31: As a peer developer I want to instantly know what properties can be without digging into the code too much. Besides Intellisense won't be really helpful when method's signature is so secretive.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,140 @@
...
+        const keys = [
+            "species",
+            "name",
+            "gender",
+            "legs",
+            "hands",
+            "paws",
+            "tail",
+            "saying",
+            "friends",
```
OleksiyRudenko, line 17: Base class shouldn't know anything about design of its children classes. This breaks SOLID principles.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,140 @@
...
+        ];
+        return keys
+            .reduce((output, key) => {
+                if (this[key]) {
+                    if (typeof this[key] === "function")
+                        return output.concat(this[key]());
+                    else return output.concat(this[key]);
+                } else return output;
+            }, [])
```
OleksiyRudenko, line 26: This can be optimized significantly. Too verbose, to much ifs. Swissknife `.reduce` doesn't add to readability when the solution can be implemented with `.map`, `.join` and other Array methods wise people in ECMAscript committee provided for us.

OleksiyRudenko (changes requested):
@AntonKottans a good start.
Please check comments and DoD bot's message below.

submissions/antonkottans/tjsw-post-oop/index.js
AntonKottans: Its a bit more verbose, but now it returns friends' names, if someone has some

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,163 @@
...
+                        else return null;
+                    else return key + ":" + this[key];
+                } else return null;
+            })
+            .filter((value) => value)
+            .join(",  ");
+    }
+
+    addFriends(...friends) {
+        if (friends)
```
OleksiyRudenko, line 30: This will always be true. No args === `[]`

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,163 @@
...
+                            return (
+                                "friends:" +
+                                this["friends"]
+                                    .map((friend) => friend.name)
+                                    .join(", ")
+                            );
+                        else return null;
+                    else return key + ":" + this[key];
+                } else return null;
+            })
```
OleksiyRudenko, line 24: Can you think of any way to simplify logic here?
As a peer developer I want methods be simpler and less nested.
Also each class should handle its own properties only.

You may benefit from methods overloading and there you won't need a property to store "keys" (which is BTW quite generic name per se).

Check the materials on OOP/SOLID in the original project's docs.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,163 @@
...
+        this.keys = [...this.keys, /* "species", */ "paws", "tail"];
+    }
+}
+
+class Human extends Creature {
+    constructor({ name, gender, friends, legs, hands, saying }) {
+        super({ name, gender, friends });
+        this.legs = legs;
+        this.hands = hands;
+        this.species = "human";
```
OleksiyRudenko, line 52: Species is a property shared by all creatures. Hence it should be defined and handled in the base class.
This is one of the points of OOP and also makes code DRY.
Generalized, shared props should be defined in a common base class.

OleksiyRudenko (changes requested):
@AntonKottans nice to start with. Yet needs significant improvements.

OleksiyRudenko:
@AntonKottans let me know if you have questions re latest review comments.

AntonKottans:
so my changes after last three comments were totally wrong?

OleksiyRudenko:
@AntonKottans 

I just noticed, the PR contains two files from similar tasks. At some point I may have read one (or both) files end-to-end to get a bigger picture and then commented on another file based on actual code and wrong bigger mental picture.

Given that I will go through `submissions/antonkottans/tjsw-post-oop/index.js` thoroughly.

Meanwhile please remove the irrelevant file as it creates confusion.

OleksiyRudenko (commented):
@AntonKottans very good overall.
Although need to create catwomen that are "synchronised" with cats. Affects all other classes and leads to making decisions and implementations that violate OOP/SOLID principles.
Please read carefully the comments below.
Generally, this type of problem is solvable with composition pattern rather than with OOP.
However, a few other options (workarounds) are possible. One of them is hinted in the comments below.
Sub-task to synchronise classes is definitely a task with added complexity.
My proposal is to build a world with no cat-women in it and no preparations to even having any, so that we have clean OOP/SOLID compliant code.
And we merge it.
Then you may want to add CatWoman class as a separate task, outside the course assignment ,and try to find a solution that affects other classes minimally.

Let me know what way you want to go further.

AntonKottans:
sorry i hope i didn't create a bigger mess this time

AntonKottans:
sorry i did revert becouse of i added fixed file to wrong directory. didnt noticed bug before pushing

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,185 @@
...
+    constructor({ name, gender, friends = [], species, saying }) {
+        this.name = name;
+        this.species = species;
+        this.gender = gender;
+        this.friends = friends;
+        this.saying = saying;
+    }
+
+    createArrayOfProps() {
+        return ["name", "species", "gender", "saying", "friends"]
```
OleksiyRudenko, line 14: This is definitely a way to list properties explicitly.
Do we still need to stick to it?
You used yet another way to list properties explicitly, i.e. `{ name, gender, friends = [], species, saying }` in method declaration. It is even better as it lists properties rather than just their names.

The following example also lists properties: `[ propA, propB, propC ]`

Destructuring laso works well on `this`, e.g. `const {propA, propB] = this` that make code DRY while yet explicit.

BTW, the following example lists props and also transforms one of them so its data type becomes similar to other props type: `[ stringA, stringB, arrayA.join(", "), stringC ]`

See how you can leverage the above to improve your code.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,185 @@
...
+                    ? "friends: " +
+                          this.friends.map((friend) => friend.name).join(", ")
+                    : "";
+            })
+            .filter((_) => _);
+    }
+
+    addFriends(...newFriends) {
+        if (newFriends.length > 0)
+            this.friends = [...this.friends, ...newFriends];
```
OleksiyRudenko, line 28: If we do not check newFriends length how it would affect outcome from `[...this.friends, ...newFriends]`?

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,185 @@
...
+        return ["name", "species", "gender", "saying", "friends"]
+            .map((key) => {
+                if (key != "friends")
+                    return key + KEYS_VALUE_SEPARATOR + this[key];
+                return this.friends.length
+                    ? "friends: " +
+                          this.friends.map((friend) => friend.name).join(", ")
+                    : "";
+            })
+            .filter((_) => _);
```
OleksiyRudenko, line 23: What filter would filter out?

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,185 @@
...
+    constructor({
+        name,
+        gender,
+        friends,
+        paws,
+        tail = "ordinary tail",
+        saying,
+    }) {
+        super({ name, gender, friends, paws, tail, species: "dog", saying });
+        this.tail = tail;
```
OleksiyRudenko, line 73: Looks like Dog and Cat have tails. Why this property is not generalized in some base class?

OleksiyRudenko (changes requested):
@AntonKottans well done!
Few improvements are still needed.
In particular proper generalization of shared props makes code DRY.

submissions/antonkottans/tjsw-post-oop/index.js
AntonKottans: if there are no friends, one of results in map's return will be false

submissions/antonkottans/tjsw-post-oop/index.js
AntonKottans: becouse not all animals has tail, like frogs

OleksiyRudenko: Do we have any tailless animals yet?
We have suboptimal code now already in favour of possible need that we do not have yet.
Pre-mature optimization is a poor practice.
Most likely if we have tailless animals we would still need split Animal class for tailed and tailless. 
So, we need a class that holds a shared property anywa.

submissions/antonkottans/tjsw-post-oop/index.js
OleksiyRudenko: Would it makes sense to strip friends **before** they are processed?

AntonKottans:
i still don't get how to gather name of property and property in certain order with one method. for now it doesn't show name of property

OleksiyRudenko:
> i still didn't get how to gather name of property and property in certain order with one method. for now it doesn't show name of property

This "Do we still need to stick to it?" was a question, not a direction to act.
If you want prop names printed, bring them back. But do this in a smart way.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  createArrayOfProps() {
+    return [...super.createArrayOfProps(), this.paws];
+  }
+}
+
+class AnimalWithTail extends Animal {
+  constructor({ name, gender, friends, paws, species, saying, tail }) {
+    super({ name, gender, friends, paws, species, saying, tail });
```
OleksiyRudenko, line 47: What `super` essentially is and what it does?

AntonKottans: its a function, that calls parent's constructor i don't need to pass in a tail to super?

AntonKottans:
no i wont return prop names, for now output inspite of legs, hands and paws is more readable.

submissions/antonkottans/tjsw-post-oop/index.js
```diff
@@ -0,0 +1,165 @@
...
+  constructor({
+    name,
+    gender,
+    friends,
+    paws,
+    tail = "ordinary tail",
+    saying = "meow!",
+  }) {
+    super({ name, gender, friends, paws, tail, species: "cat", saying });
+    this.saying = saying;
```
OleksiyRudenko, line 84: Any good reasons to initialize saying two ways?

OleksiyRudenko (approved):
@AntonKottans great job done!

See the comment below. Let's have it as a learning to check your code for bugs.

submissions/antonkottans/tjsw-post-oop/index.js
AntonKottans: i added this.saying = saying; for catwoman. there is no need for that

OleksiyRudenko: @AntonKottans You are doing good and will be fine. There is no learning without an effort.

PR #123, opened 2022-08-11
Hooli-style Popup

Yanyshpolska:
### Hooli-style Popup

[Demo](https://yanyshpolska.github.io/HWPopup/) |
[Code base](https://github.com/Yanyshpolska/HWPopup)

Please, review.

### DIFF
--- a/submissions/yanyshpolska/Hooli-style Popup/css/style.css
+++ b/submissions/yanyshpolska/Hooli-style Popup/css/style.css
@@ -41,7 +41,7 @@ a {
 }
 .menu__item_text {
   border: 0.15rem solid transparent;
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   color: #555;
   padding: 0.2rem 0.15rem;
 }
@@ -54,7 +54,7 @@ a {
 .menu__item_picture {
   width: 1.715rem;
   border: 0.15rem solid transparent;
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   padding: 0.15rem;
 }
 .menu__item_picture_img {
@@ -78,7 +78,7 @@ a {
   height: 1.715rem;
   width: 1.715rem;
   border: 0.15rem solid transparent;
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   background: url("../img/popup-button.png") 50% 50%/80% 80% no-repeat;
 }
 .visually-hidden {
@@ -106,7 +106,7 @@ a {
   right: -6.45rem;
   top: 2.15rem;
   border: 0.15rem solid rgb(135, 105, 145);
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   background-color: #fffaf2;
   width: 22rem;
   height: 23.5rem;
@@ -121,7 +121,7 @@ a {
   user-select: none;
   flex: 0 0 30%;
   border: 0.15rem solid transparent;
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   margin: 0.33rem;
   padding: 0.3rem 0rem;
   display: flex;
@@ -158,7 +158,7 @@ a {
   justify-content: center;
   padding: 0.15rem;
   background-color: #fde4ba;
-  border-radius: 0.2rem;
+  border-radius: 0.2em;
   width: 18rem;
   border: 0.15rem solid transparent;
 }
--- a/submissions/yanyshpolska/Hooli-style Popup/css/style.css
+++ b/submissions/yanyshpolska/Hooli-style Popup/css/style.css
@@ -177,6 +177,7 @@ ul li {
 }
 .popup__item_checkbox_input:checked ~ .popup__item_more {
   display: block;
+  border-top: 1px solid #876991;
 }
 .popup__item_checkbox_label {
   margin: auto;
@@ -192,14 +193,9 @@ ul li {
   display: block;
   color: #555;
 }
-.popup__item_checkbox_less {
+
+.popup__item_checkbox_input:checked ~ .popup__item_checkbox_label {
   display: none;
-  color: #555;
-}
-.popup__item_checkbox_input:checked
-  ~ .popup__item_checkbox_label
-  > .popup__item_checkbox_less {
-  display: block;
 }
 .popup__item_checkbox_input:checked
   ~ .popup__item_checkbox_label
--- a/submissions/yanyshpolska/Hooli-style Popup/css/style.css
+++ b/submissions/yanyshpolska/Hooli-style Popup/css/style.css
@@ -170,6 +170,9 @@ a {
 .popup__item_checkbox_input:checked ~ .popup__item_checkbox_label {
   display: none;
 }
+.popup__item_checkbox_input:checked {
+  display: none;
+}
 .popup__item_checkbox_input:checked
   ~ .popup__item_checkbox_label
   > .popup__item_checkbox_more {
--- a/submissions/yanyshpolska/Hooli-style Popup/css/style.css
+++ b/submissions/yanyshpolska/Hooli-style Popup/css/style.css
@@ -14,18 +14,11 @@ html,
 body {
   height: 100%;
   width: 100%;
-  font-size: 100%;
   line-height: 1;
   font-size: 14px;
-  -ms-text-size-adjust: 100%;
-  -moz-text-size-adjust: 100%;
-  -webkit-text-size-adjust: 100%;
 }
 a {
   text-decoration: none;
-}
-ul li {
-  list-style: none;
 }
 .main {
   height: 100%;
@@ -81,6 +74,7 @@ ul li {
   min-width: 1.715em;
 }
 .menu__item_checkbox_label {
+  cursor: pointer;
   position: absolute;
   height: 1.715em;
   width: 1.715em;
@@ -159,6 +153,7 @@ ul li {
   border-top: 1px solid #876991;
 }
 .popup__item_checkbox_label {
+  cursor: pointer;
   margin: auto;
   display: flex;
   justify-content: center;
--- a/submissions/yanyshpolska/Hooli-style Popup/css/style.css
+++ b/submissions/yanyshpolska/Hooli-style Popup/css/style.css
@@ -1,12 +1,8 @@
-/* Nullable style */
+/* base styles */
 * {
   padding: 0;
   margin: 0;
   border: 0;
-}
-*,
-*:before,
-*:after {
   -moz-box-sizing: border-box;
   -webkit-box-sizing: border-box;
   box-sizing: border-box;
@@ -14,16 +10,6 @@
 :focus,
 :active {
   outline: none;
-}
-a:focus,
-a:active {
-  outline: none;
-}
-nav,
-footer,
-header,
-aside {
-  display: block;
 }
 html,
 body {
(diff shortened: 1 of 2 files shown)


kasionio:
Please try to reach `Less` button from the keyboard, now it looks impossible :)

Yanyshpolska:
Hi! Last hour I have tried to solve this problem. The button is actually "more" button, so you can reach it on position between "always visible items" and "more items" items. And it indeed seems weird than you must press "tab" back to focus on the "Less" button. The "~" and "+" (as I understand, and as I tried) can reach only the next element in the document. If I add new checkbox button to the end of item list - it will nor reach upper element.
So, I have tried add radio button  with "More" and "Less" items. But it behaved in the same manner as with the checkbox, so this solution didn't work here for me.
Also I have tried to set order of flex element for checkbox to sent it back, but it didn't change the real order of objects, so in it also didn't work.
I know this can be done with JS but I didn't find solution how to do it purely in HTML/CSS only.
Please point me in the right direction :)

Yanyshpolska:
Hi! As I understand, "alt" descriptions for icons in the popup menu are not necessary, so  I've deleted them. I'm still waiting for advice about the "Less" button. Or I can remove it and hide the "More" button when pressing it.

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,213 @@
+/* base styles */
```
kasionio, line 1: please remove comments, they are not needed, code should be readable by its nature :D

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,213 @@
...
+* {
```
kasionio, line 2: 👍

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,213 @@
...
+  align-items: center;
+}
+.menu__item {
+  display: block;
+  height: 1.715em;
+  margin: 0.715em;
+}
+/* ===============Menu text items======================= */
+.menu__item_text {
+  /* display: block; */
```
kasionio, line 55: all commented out code should be deleted

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,213 @@
...
+  border-color: #c4b4c9;
+}
+/* ===============Menu checkbox item===================*/
+.menu__item_checkbox {
+  position: relative;
+  min-width: 1.715em;
+}
+.menu__item_checkbox_label {
+  position: absolute;
+  height: 1.715em;
```
kasionio, line 94: Try to use also rem :) 
https://essential-addons.com/elementor/css-design-guide-difference/
https://www.quora.com/What-is-the-difference-between-em-rem-and-px-in-CSS

submissions/yanyshpolska/Hooli-style Popup/index.html
```diff
@@ -0,0 +1,220 @@
...
+              </label>
+            </div>
+          </div>
+        </div>
+
+        <a class="menu__item_picture menu__item" href="#">
+          <img
+            class="menu__item_picture_img"
+            src="./img/bell.png"
+            alt="Alerts"
```
kasionio, line 210: https://www.a11yproject.com/posts/alt-text/

submissions/yanyshpolska/Hooli-style Popup/index.html
```diff
@@ -0,0 +1,220 @@
...
+    <link rel="icon" type="image/x-icon" href="/img/favicon.png" />
+    <link rel="stylesheet" href="css/style.css" />
+    <title>Hooli-style Popup</title>
+  </head>
+  <body>
+    <header>
+      <nav class="menu__items">
+        <a class="menu__item_text menu__item" href="#">Mail</a>
+        <a class="menu__item_text menu__item" href="#">Images</a>
+
```
kasionio, line 16: it is not needed to leave spaces in HTML file

submissions/yanyshpolska/Hooli-style Popup/index.html
```diff
@@ -0,0 +1,220 @@
...
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="icon" type="image/x-icon" href="/img/favicon.png" />
+    <link rel="stylesheet" href="css/style.css" />
+    <title>Hooli-style Popup</title>
+  </head>
+  <body>
+    <header>
+      <nav class="menu__items">
```
kasionio, line 13: for plural naming better use `list` word for example, because item and items looks pretty the same and you need to pay much attention to see the difference

submissions/yanyshpolska/Hooli-style Popup/index.html
```diff
@@ -0,0 +1,220 @@
...
+            id="menu_checkbox"
+            class="menu__item_checkbox_input visually-hidden"
+            type="checkbox"
+          />
+          <label for="menu_checkbox" class="menu__item_checkbox_label"> </label>
+
+          <div class="popup__wrapper">
+            <div class="popup__items">
+              <a href="#" class="popup__item">
+                <div class="popup__item_image_container">
```
kasionio, line 28: you can try to remove this div or text span and reduce the amount of tags, or wrap text and img in one div and use `flex`.

kasionio (changes requested):
Nice job, good code! Let's improve it a bit. Take a look at the comments and screenshot. Also, need to add a pointer on hover effect on the popup button and More button. Focus on popup button is not centered
image

submissions/yanyshpolska/Hooli-style Popup/css/style.css
Yanyshpolska: Agreed. My bad. I'll delete them.

submissions/yanyshpolska/Hooli-style Popup/index.html
Yanyshpolska: For this particular case, I believe that usage of the "alt" is reasonable: in case the picture is not loaded the user is still able to see picture caption which is aligned with advice from this article: 
"Use the alt attribute for images that are used as more than decoration."

submissions/yanyshpolska/Hooli-style Popup/index.html
Yanyshpolska: Here I use div-container-for-picture in purpose because all pictures have different sizes and this approach allows me to deal with it: regardless of the size of the picture, the menu item cells will be the same.

Yanyshpolska:
Please check the code adjusted according to your comments. I have corrected the focus on the popup button and deleted extra spaces and all comments. Also, I have changed class names.

submissions/yanyshpolska/Hooli-style Popup/index.html
kasionio: You can specify image width explicitly in corresponding class for images. You can leave it, but this is far from optimal

submissions/yanyshpolska/Hooli-style Popup/index.html
kasionio: Yes, it is reasonable, but the value is not correct

Yanyshpolska:
Please check the code adjusted according to your comments. I have removed "div-container-for-picture" and changed the "alt" value from "Alerts" to "Notifications".

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,191 @@
...
+  outline: none;
+}
+html,
+body {
+  height: 100%;
+  width: 100%;
+  font-size: 100%;
+  line-height: 1;
+  font-size: 14px;
+  -ms-text-size-adjust: 100%;
```
kasionio, line 20: what does this property mean?

submissions/yanyshpolska/Hooli-style Popup/css/style.css
```diff
@@ -0,0 +1,191 @@
...
+  -moz-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+}
+a {
+  text-decoration: none;
+}
+ul li {
```
kasionio, line 27: you can leave just ul or li here

submissions/yanyshpolska/Hooli-style Popup/index.html
```diff
@@ -0,0 +1,182 @@
...
+        </div>
+        <a class="menu__item_picture menu__item" href="#">
+          <img
+            class="menu__item_picture_img"
+            src="./img/bell.png"
+            alt="Notifications"
+          />
+        </a>
+        <a class="menu__item_picture menu__item" href="#">
+          <img class="menu__item_picture_img" src="./img/logo.jpg" alt="User" />
```
kasionio, line 176: https://www.w3schools.com/tags/att_img_alt.asp

kasionio (changes requested):
Please add the hover pointer effect on popup button and More button

Yanyshpolska:
Added hover pointer and removed unnecessary rows.

submissions/yanyshpolska/Hooli-style Popup/index.html
Yanyshpolska: Changed the "alt" value from "User" to "Profile".

kasionio: `The text should explain where the link goes if the image is inside an <a> element` 
So, it should be "Go to Profile page"

submissions/yanyshpolska/Hooli-style Popup/css/style.css
kasionio: Please read the articles (these or any others) and add at least a couple of rems, to the corresponding classes. `Rem` is more used nowadays than `em`

kasionio (changes requested):
Last steps ;)

submissions/yanyshpolska/Hooli-style Popup/index.html
kasionio: Made corresponding changes to `alt`s

Yanyshpolska:
Hi! Thanks for your comments. Reworked as advised. Please review.

kasionio (changes requested):
nope, please read the articles and add rems where you think it is NEEDED, not everywhere.

Yanyshpolska:
Hi Illia. Of course, I read the articles which you suggested to me. I also read other articles and watched some videos about units of measure. I clearly understand the differences in EM and REM size units. But I don't know the "best practice" of using them. I believe that the experience of use will come with time and with the number of implemented projects. 
This popup menu project is my first project. I have used EM units here only to have some practice with them. The first version of popup menu was written with the PX. In my CSS file I set the font size only one time only for the HTML tag. So I expect the behavior of EM and REM will be the same. 
If you have some suggestions for "best practice" solutions with EM REM - I will gladly listen to them. But for now, I can only guess where to use EM or REM units better in this particular project and I did the best I can considering my current knowledge.
According to the articles I read (https://zellwk.com/blog/rem-vs-em/, https://zellwk.com/blog/media-query-units/) I have found some suggestions for using EM and REM: "Size in em if the property scales according to its font-size", "Size everything else in rem." "Use em for media queries." According to your article, I have changed "border-radius" to EM.

kasionio:
Nice answer, fair enough ;) I guess it is time to go to the next tasks 🚀

kasionio (approved):

PR #128, opened 2022-08-12
memory pair game

AntonKottans:
Memory pair game

[Demo](https://antonkottans.github.io/memory-pair-game/) |
[Code base](https://github.com/AntonKottans/memory-pair-game)

Please, review.


submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
...
+        };
+        timerId = setInterval(() => {
+            increaseTimer();
+            if (minutes === 60) timerOutput.textContent = "you lost";
+            timerOutput.textContent = getFormattedTimer();
+        }, 1000);
+    };
+
+    const createClickHandler = () => {
+        let openned = 0;
```
artem-trubin, line 29: By storing this value into a closure and not some global scope or a state, you introduced a bug, where someone can open 5 pairs of cards, press "Start" to restart the game, open 1 pair and win, since you're not checking for all open pairs, you're checking for a certain amount of pairs. Either clean this counter on game restart or make a better win condition check. Either way, avoid using closures in such way, it complicates the code and invites mistakes. 

Also `opened` is a name for variable of a boolean type. Try asking your variable name as a question and you'll see if it fits the value type.

_opened? - true_ seems appropriate.
_opened? - 2_ seems confusing.
_openedAmount? - 2_ seems a lot better.

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
...
+        let flipped = undefined;
```
artem-trubin, line 30: Check comment above, this could be `false` by default.

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
...
+                        target.closest('*[class*="card_"]').children[1];
+
+                    if (!flipped) {
+                        flipped = {
+                            container: currentCardContainer,
+                            face: currentCardFace,
+                        };
+                    } else {
+                        if (
+                            flipped.face.getAttribute("src").slice(9, -4) ===
```
artem-trubin, line 54: As I can see here, you're keeping information about the state of the game (which cards are open and which are closed) right in the DOM elements. This is wrong, because DOM elements are easily editable from the browser element inspector, so a user can cheat in your game, manually writing `opened` or other classes in these elements. 

You should find a way to store such information in JS and make DOM only _represent_ this information visually. DOM should __NEVER__ be the source of data except when you explicitly wait some information from a user, like in a login form. And you should __NEVER__ store internal information in DOM elements.

Hint: you could use an array of `true`/`false` values representing each card and when some card is flipped, get it's id and connect it to the number of `true`/`false` value in this array. You'll keep data about cards state in one place, not reachable to user. Also, win condition check writes itself this way - just check for the whole array to be `true`.

submissions/antonkottans/memory-pair-game/style.css
```diff
@@ -0,0 +1,93 @@
...
+    padding: 20px;
+}
+
+.header {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 100%;
+}
+.result {
```
artem-trubin, line 15: Be consistant with empty lines, place one between each block for better readability.

artem-trubin (commented):
Hi! Let's make some changes to improve your work.

1. Add CSS styles that forbid users to select or drag images in your cards. To avoid behaviour like this: 
image
2. What are `forgger` and `tjsw-post-oop` folders are doing in this PR? Please leave only files that are related to the Memory Pair Game task.
3. There are a lot of comments for the code, so let's summarize what you'll have to do for your next review in short thesis':
- if you can pass an argument to function instead of relying on global variable - do it
- create functions outside function if possible to avoid taking memory space by instancing similar functions
- don't store data in the DOM elements. DOM should represent data from JS
- almost everything here could be `const`. Use it by default and switch to `let` only if you're 100% sure the variable will be changed (TIP: changing array with `.push` and other methods and changing object properties is _NOT_ changing the variable. All of this is possible with arrays and objects being constants.

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
+document.addEventListener("DOMContentLoaded", () => {
+    let timerId = undefined;
+
+    const startTimer = () => {
```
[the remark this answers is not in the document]
AntonKottans: i thought it would be more readable to nest function in its using space

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
...
+                }, 1000);
+            }
+        };
+    };
+    const startGame = () => {
+        clearInterval(timerId);
+        document
+            .querySelectorAll(".opened")
+            .forEach((elem) => elem.classList.remove("opened"));
+        setTimeout(() => {
```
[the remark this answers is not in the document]
AntonKottans: setTimeout sets delay for animation

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,206 @@
...
+        }, 1000);
+    };
+    const stopGame = (result = "win") => {
+        clearInterval(timerId);
+        let span = document.createElement("span");
+        span.classList.add("result");
+        span.textContent =
+            result === "win"
+                ? `You won! You time:${
+                      document.querySelector(".timer").textContent
```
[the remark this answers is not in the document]
AntonKottans: to get node and apply changes to him, i guess

submissions/antonkottans/memory-pair-game/script.js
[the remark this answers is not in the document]
artem-trubin: 1. You basically nested your whole login inside a function and it gave the whole file one extra indentation.
2. When you nest function A in function B, every time function B is called, an instance of function A is created in the memory. So there could be a lot of instances of A if B would be called a lot of times. And you should not rely on the garbage collectors to clear those unused instances.

submissions/antonkottans/memory-pair-game/script.js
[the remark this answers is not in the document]
artem-trubin: I don't think you understood the comment here. You are trying to get your "time" score _FROM_ the DOM. You should not rely on data being in DOM. You already have your _time_ score in JS code, use it.

artem-trubin:
I understand that it's convenient to use "Resolve conversation" as some kind of to-do list checker, but please don't resolve conversations that are not yet "fixed" (you didn't push code) and especially when you leave comments in them. It looks very confusing for the reviewer to get a message about a comment and then search it somewhere among the resolved conversations in the PR.

AntonKottans:
> I understand that it's convenient to use "Resolve conversation" as some kind of to-do list checker, but please don't resolve conversations that are not yet "fixed" (you didn't push code) and especially when you leave comments in them. It looks very confusing for the reviewer to get a message about a comment and then search it somewhere among the resolved conversations in the PR.

sorry, i will not be doing like this

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
+const pipe = (...arrayOfFunctions) => {
```
artem-trubin, line 1: Is there a reason for `pipe` usage besides just to use pipe? It is used only once and makes code much more complicated.

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+    if (arrayOfFunctions.length === 0) return undefined;
+    else if (arrayOfFunctions.length === 1) return arrayOfFunctions[0]();
+    return arrayOfFunctions.reduce((prev, func) => {
+        return func(prev);
+    }, undefined);
+};
+let timerId = null;
```
artem-trubin, line 8: Why use a global variable if `timerId` is in the state?

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+        state.cardContainers = Array.from(
+            document.querySelectorAll(".card-container")
+        );
+    state.cardContainers.forEach((node) => {
+        if (
+            state.guessedIds.has(
+                node
+                    .getAttribute("class")
+                    .split(" ")
+                    .find((clas) => {
```
artem-trubin, line 59: `class`

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+};
+let timerId = null;
+
+const state = {
+    timerId: null,
+    guessedPairsAmount: 0,
+    flipped: false,
+    guessedPairs: [],
+    guessedIds: new Set(),
+    notGuessedIds: new Set(),
```
artem-trubin, line 16: What is the point of this set?

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+    }, undefined);
+};
+let timerId = null;
+
+const state = {
+    timerId: null,
+    guessedPairsAmount: 0,
+    flipped: false,
+    guessedPairs: [],
+    guessedIds: new Set(),
```
artem-trubin, line 15: Why `guessedPairds` and `cardContainers` are arrays and `guessedIds` and `notGuessedIds` are sets?

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+    return arrayOfFunctions.reduce((prev, func) => {
+        return func(prev);
+    }, undefined);
+};
+let timerId = null;
+
+const state = {
+    timerId: null,
+    guessedPairsAmount: 0,
+    flipped: false,
```
artem-trubin, line 13: This value is defined as `boolean` here, but later in the code objects are placed here. If you want some placeholder that will mean "nothing is here yet", use `null`.

submissions/antonkottans/memory-pair-game/script.js
```diff
@@ -0,0 +1,250 @@
...
+    state.bodyNode.appendChild(span);
+    span.addEventListener("click", () => {
+        state.bodyNode.removeChild(span);
+    });
+};
+
+const getDoubledRandomImageNodes = (amountOfElements) => {
+    if (amountOfElements <= 0) return [];
+
+    const getCardNames = (amount) => {
```
artem-trubin, line 152: This function doesn't return Card Names. It's misleading. It returns another function that, if called, returns card names. I believe you've made it that complicated because of pipe. There is no reason for it.

artem-trubin (commented):
Hey! There are some more improvements that should be made. Check the comments below.

submissions/antonkottans/memory-pair-game/script.js
AntonKottans: Becouse     
return pipe(
        getCardNames(amountOfElements),
        doubleAndShuffle,
        convertNamesToNodes
);
is more readable then
return convertNamesToNodes(doubleAndShuffle(getCardNames(amountOfElements)))

artem-trubin (approved):
Good job!

PR #135, opened 2022-08-12
frogger task

the-darkesttt:
### frogger game task

[Demo](https://the-darkesttt.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/the-darkesttt/frontend-nanodegree-arcade-game)

I'm adding into PR only app.js though it's not the only one I've changed in this project.

Please, review.

### DIFF
--- a/submissions/the-darkesttt/frogger-task/app.js
+++ b/submissions/the-darkesttt/frogger-task/app.js
@@ -40,14 +40,14 @@ class Menu {
 
         playerSkins.forEach((skin) => {
             const menuItem = document.createElement('li');
+            menuItem.classList.add('menu-item');
+            menuItem.dataset.skin = skin.id;
     
             const itemImg = document.createElement('img');
             const imgSrc = skin.src;
-            const imgId = skin.id;
             itemImg.setAttribute('src', imgSrc);
-            itemImg.setAttribute('id', imgId);
             const imgDesc = document.createElement('p');
-            imgDesc.innerText = imgId;
+            imgDesc.innerText = skin.id;
     
             this.menuList.appendChild(menuItem);
             menuItem.appendChild(itemImg);
@@ -65,7 +65,7 @@ class Menu {
         this.menuList.addEventListener('click', ({ target }) => {
             if (target.tagName === "UL") return;
             
-            const targetId = target.getAttribute("id");
+            const targetId = target.closest('.menu-item').dataset.skin;
 
             const targetSprite = playerSkins.find((skin) => {
                 if (skin.id === targetId) {
--- a/submissions/the-darkesttt/frogger-task/app.js
+++ b/submissions/the-darkesttt/frogger-task/app.js
@@ -22,16 +22,6 @@ const playerSkins = [
 ];
 
 class Menu {
-    bg;
-    menu;
-    menuTitle;
-    menuDescriprion;
-
-    menuList;
-    menuItem;
-
-    menuScore;
-    menuClose;
 
     constructor() {
         this.bg = document.createElement('div');
@@ -48,20 +38,20 @@ class Menu {
         this.menuDescriprion.innerText = 'Choose your character!';
         this.menuList = document.createElement('ul');
 
-        playerSkins.forEach((item) => {
-            this.menuItem = document.createElement('li');
+        playerSkins.forEach((skin) => {
+            const menuItem = document.createElement('li');
     
             const itemImg = document.createElement('img');
-            const imgSrc = item.src;
-            const imgId = item.id;
+            const imgSrc = skin.src;
+            const imgId = skin.id;
             itemImg.setAttribute('src', imgSrc);
             itemImg.setAttribute('id', imgId);
             const imgDesc = document.createElement('p');
             imgDesc.innerText = imgId;
     
-            this.menuList.appendChild(this.menuItem);
-            this.menuItem.appendChild(itemImg);
-            this.menuItem.appendChild(imgDesc);
+            this.menuList.appendChild(menuItem);
+            menuItem.appendChild(itemImg);
+            menuItem.appendChild(imgDesc);
             
         });
 
@@ -72,11 +62,18 @@ class Menu {
         this.menu.appendChild(this.menuDescriprion);
         this.menu.appendChild(this.menuList);
 
-        this.menuList.addEventListener('click', (event) => {
-            const listTarget = event.target;
-            if (listTarget.tagName == "UL") return;
-        
-            player.sprite = listTarget.getAttribute("src");
+        this.menuList.addEventListener('click', ({ target }) => {
+            if (target.tagName === "UL") return;
+            
+            const targetId = target.getAttribute("id");
+
+            const targetSprite = playerSkins.find((skin) => {
+                if (skin.id === targetId) {
+                    return skin;
+                }
+            });
+
+            player.sprite = targetSprite.src;
     
             this.bg.classList.add('menu-animation');
             this.menu.classList.add('menu-animation');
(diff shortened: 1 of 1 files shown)


submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        src: 'images/char-pink-girl.png',
+    },
+    {
+        id: 'princess-girl',
+        src: 'images/char-princess-girl.png',
+    },
+];
+
+class Menu {
+    bg;
```
artem-trubin, line 25: It is not required to define properties like that before the constructor. They will be created there.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        this.menuTitle.innerText = 'Game Menu';
+        this.menuDescriprion = document.createElement('p');
+    }
+
+    start() {
+
+        this.menuDescriprion.innerText = 'Choose your character!';
+        this.menuList = document.createElement('ul');
+
+        playerSkins.forEach((item) => {
```
artem-trubin, line 51: You can give more distinctive names to arguments and variables. Like, for example, here you can swap `item` with `skin`. `skin.src` and `skin.id` would give more sense this way, right?

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+            this.menuItem = document.createElement('li');
```
artem-trubin, line 52: Is there a reason for this to be a property of the `Menu` class? Looks like could be just a normal constant variable.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        document.body.appendChild(this.bg);
+        document.body.appendChild(this.menu);
+
+        this.menu.appendChild(this.menuTitle);
+        this.menu.appendChild(this.menuDescriprion);
+        this.menu.appendChild(this.menuList);
+
+        this.menuList.addEventListener('click', (event) => {
+            const listTarget = event.target;
+            if (listTarget.tagName == "UL") return;
```
artem-trubin, line 77: Always use strict comparisons (`===`).

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        
+            player.sprite = listTarget.getAttribute("src");
```
artem-trubin, line 79: Here it is fine since we are still learning and the skin picker wasn't even part of the main task, but keep this in mind:
_NEVER_ rely on something being in DOM. Here, for example, getting a path to the skin from the `src` property seems like a neat shortcut, but remember that the user can easily open dev tools in the browser, edit the `src` in the element and it would break everything since there is no such path in your files. This exact code piece could be implemented by giving each "skin button" an ID and then, within JS you could connect each ID to some skin path. Actually, you already have an array like that at the start of the file.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        });
+
+        document.body.appendChild(this.bg);
+        document.body.appendChild(this.menu);
+
+        this.menu.appendChild(this.menuTitle);
+        this.menu.appendChild(this.menuDescriprion);
+        this.menu.appendChild(this.menuList);
+
+        this.menuList.addEventListener('click', (event) => {
```
artem-trubin, line 75: Since you only use `target` property from the `event` class, you can get it by desctructuring here:
```js
this.menuList.addEventListener('click', ({ target }) => {
```

It works exactly like
```js
this.menuList.addEventListener('click', (event) => {
    const target = event.target
```

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        });
+    }
+}
+
+
+const startMenu = new Menu();
+startMenu.start();
+
+let points = 0;
+const score = document.createElement('p');
```
artem-trubin, line 124: Better names: `score` kinda sounds like `points`, though it doesn't store value, it stores an element. Maybe `scoreElem` could be better.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+updateScore();
+
+function animateScoreGreen() {
+    score.classList.add('green');
+    setTimeout(function() {
+        score.classList.remove('green');
+    }, 1000)
+}
+
+function animateScoreBlue() {
```
artem-trubin, line 141: Functions `animateScoreBlue` and `animeateScoreGreen` are identical, the only thing changing here is the color. You can write a neat tool for yourself here to highlight the scoreboard with some color by creating a function that would receive a color and a timeout value and just reuse it. Like 
```js
function hightlightScore(color, duration) { ...
```

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+}
+
+class Enemy extends Entity {
+    speed = 100 + Math.floor(Math.random() * 222);
+
+    constructor(x, y, sprite) {
+        super(x, y, sprite);
+    }
+
+    handleInput(){}
```
artem-trubin, line 221: Does `Enemy` need a `handleInput` method?

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+
+    update(dt) {
+
+        if (this.sprite == "images/enemy-bug.png") {
```
artem-trubin, line 225: Use strict comparisons

artem-trubin, line 225: It feels wrong to make this condition based on the sprite type. It would be better to add a property to enemies, like `facingDirection`, that would just be left/right.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+            this.x += this.speed * dt;
+
+            if (this.x > 500) {
```
artem-trubin, line 228: There are a lot of magic numbers here, `500`, `222`, `70`, `60`. For a reviewer (or even you in ten years) they don't make sense. Maybe it's better to set them to constants to make the code more readable.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    if (healthBar.children.length > 0){
+        let imgArr = Array(...healthBar.children);
+        imgArr.forEach(item => {
+            item.remove();
+        });
+    }
+
+    for (let i = 1; i <= health; i++) {
```
artem-trubin, line 167: Better go 
```js
for (let i = 0; i < health; i++) {...
```
When you use for and starting `i` doesn't equal `0` it draws attention to why is that. I believe because you had 4 hearts instead of 3 you are starting with `1` instead of `0`. Your problem was that `<=`. Just use `<` and `i` will be `0, 1, 2` and not `0, 1, 2, 3`.

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        let img = document.createElement('img');
```
artem-trubin, line 168: `const`?

submissions/the-darkesttt/frogger-task/app.js
```diff
@@ -0,0 +1,323 @@
...
+        super(x, y, sprite);
+    }
+
+    handleInput(keyPress) {
+
+        if (keyPress == 'left' && this.x > 0) {
+            this.x -= blockSize.width;
+        };
+    
+        if (keyPress == 'right' && this.x < 400) {
```
artem-trubin, line 275: More magic numbers, isn't this `400` connected to `canvasSize`?

artem-trubin (commented):
Great work and bonus points for creativity with skin selector! Let's make it even better. Check comments below.

artem-trubin:
Hey, nice improvements!

There is currently a bug, to reproduce:
1. Start the game
2. On the Skin Selector screen click not the image of the skin, but its name. 
3. An error is dropped in the console.

Judging by the text of the error, I could assume that the problem is that you get `src` for the skin directly from the clicked image of the skin, and when someone clicks on the button, but on text instead of image, it breaks, since text has no `src`. You need to figure out some sort of button-to-src connection for the whole button to work properly.

the-darkesttt:
@artem-trubin please review

artem-trubin:
Good job!

PR #138, opened 2022-08-13
memory-pair-game

Eugene-CG:
### Memory Pair Game

[Demo](https://eugene-cg.github.io/Memory-Pair-Game/)
[Code base](https://github.com/Eugene-CG/Memory-Pair-Game)

Please, review.

### DIFF
--- a/submissions/Eugene-CG/memory-pair-game/index.js
+++ b/submissions/Eugene-CG/memory-pair-game/index.js
@@ -1,18 +1,6 @@
 const cardsList = document.querySelector(".cards__list");
-let clickRemember = [];
-class Card {
-  constructor(img) {
-    this.img = img;
-  }
-  createCard() {
-    return `
-    <li class="card__container">
-          <div class="card card-back">
-              <img src="./img/${this.img}">
-          </div>
-      </li>`;
-  }
-}
+let clicksStorage = [];
+
 const startGame = () => {
   let cards = [
     "fool.jpg",
@@ -24,23 +12,30 @@ const startGame = () => {
     "hanged.jpg",
     "justice.jpg",
   ];
-  cards = [...cards, ...cards];
-  shuffleCards(cards);
-  cards = createCardOjects(cards);
-  createAllCards(cards);
+  cards = shuffleCards([...cards, ...cards]);
+  let cardListInnerHtml = ``;
+  cards.forEach((imgSrc) => {
+    cardListInnerHtml += createCard(imgSrc);
+  });
+  cardsList.innerHTML = cardListInnerHtml;
 };
 const shuffleCards = (cards) => cards.sort(() => Math.random() - 0.5);
-const createCardOjects = (cards) => cards.map((imgSrc) => new Card(imgSrc));
-const createAllCards = (cards) =>
-  cards.forEach((card) => (cardsList.innerHTML += card.createCard()));
+const createCard = (imgSrc) => {
+  return `
+    <li class="card__container">
+          <div class="card card-back">
+              <img src="./img/${imgSrc}">
+          </div>
+      </li>`;
+};
+const flipCard = (target) => {
+  //****************************** * * * * ***/
+  if (target.closest(".card")) {
+    clicksStorage = [...clicksStorage, target.firstElementChild];
+    target.classList.remove("card-back");
+    target.classList.add("hide", "open");
 
-const flipCard = ({ target }) => {
-  if (target.closest(".card") || target.closest("img")) {
-    clickRemember = [...clickRemember, target.firstElementChild];
-    target.closest(".card").classList.remove("card-back");
-    target.closest(".card").classList.add("hide", "open");
-
-    if (clickRemember.length === 2) {
+    if (clicksStorage.length === 2) {
       setTimeout(() => {
         checkMatch();
       }, 1000);
@@ -50,22 +45,22 @@ const flipCard = ({ target }) => {
 const checkMatch = () => {
   if (
     !(
-      clickRemember[0].getAttribute("src") ===
-      clickRemember[1].getAttribute("src")
+      clicksStorage[0].getAttribute("src") ===
+      clicksStorage[1].getAttribute("src")
     )
   )
     checkUnmatch();
-  clickRemember = [];
+  clicksStorage = [];
 };
 const checkUnmatch = () => {
-  clickRemember.forEach((target) => {
+  clicksStorage.forEach((target) => {
     target.closest(".card").classList.add("card-back");
     target.closest(".card").classList.remove("hide", "open");
   });
-  clickRemember = [];
+  clicksStorage = [];
 };
 
 cardsList.addEventListener("click", (event) => {
-  if (!(clickRemember.length === 2)) flipCard(event);
+  if (!(clicksStorage.length === 2)) flipCard(event.target);
 });
 window.addEventListener("DOMContentLoaded", startGame);
--- a/submissions/Eugene-CG/memory-pair-game/index.js
+++ b/submissions/Eugene-CG/memory-pair-game/index.js
@@ -4,53 +4,54 @@ let clicksStorage = [];
 const startGame = () => {
   let cards = [
     {
-      imgClass: "fool",
+      imgId: "fool",
       src: "fool.jpg",
     },
     {
-      imgClass: "lovers",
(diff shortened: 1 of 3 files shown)


submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+  createCard() {
+    return `
+    <li class="card__container">
+          <div class="card card-back">
+              <img src="./img/${this.img}">
+          </div>
+      </li>`;
+  }
+}
```
zonzujiro, line 15: No need for this `class` here. There is no inheritance, no logic inside, etc. Basically, you're using it for data, so just use objects :)

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,71 @@
...
+  cards = createCardOjects(cards);
+  createAllCards(cards);
+};
+const shuffleCards = (cards) => cards.sort(() => Math.random() - 0.5);
+const createCardOjects = (cards) => cards.map((imgSrc) => new Card(imgSrc));
+const createAllCards = (cards) =>
+  cards.forEach((card) => (cardsList.innerHTML += card.createCard()));
+
+const flipCard = ({ target }) => {
+  if (target.closest(".card") || target.closest("img")) {
```
zonzujiro, line 38: Why this `||`? This two elements has one parent, right?

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,71 @@
...
+    clickRemember = [...clickRemember, target.firstElementChild];
+    target.closest(".card").classList.remove("card-back");
```
zonzujiro, line 40: Instead of running search in DOM for every card, you can save it into variable inside listener

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,71 @@
+const cardsList = document.querySelector(".cards__list");
+let clickRemember = [];
```
zonzujiro, line 2: Just try to translate this in google translate :) And imagine how it will look like for native speaker :)

https://www.grammarly.com/

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,71 @@
...
+  ];
+  cards = [...cards, ...cards];
+  shuffleCards(cards);
+  cards = createCardOjects(cards);
+  createAllCards(cards);
+};
+const shuffleCards = (cards) => cards.sort(() => Math.random() - 0.5);
+const createCardOjects = (cards) => cards.map((imgSrc) => new Card(imgSrc));
+const createAllCards = (cards) =>
+  cards.forEach((card) => (cardsList.innerHTML += card.createCard()));
```
zonzujiro, line 35: DOM manipulation in loops.

Adding elements to DOM from a loop is a bad practice. A browser will run reflow and repaint for every element in the loop. Instead, you can:

Use append method, which can add several elements in one operation
Create some wrapper, add your items to the wrapper and then add it to DOM. It will be one operation.
Clone current container. Add items to a container and then replace your old container with a new one. But be aware of event listeners.
Use innerHTML instead

zonzujiro (changes requested):

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+  ];
+  cards = shuffleCards([...cards, ...cards]);
+
+  let cardListInnerHtml = ``;
+  let i = 0;
+  cards.forEach((obj) => {
+    cardListInnerHtml += createCard(obj, i++);
+  });
+  cardsList.innerHTML = cardListInnerHtml;
```
zonzujiro, line 46: `Array#forEach` callback has second argument - `index`

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+};
+const shuffleCards = (cards) => cards.sort(() => Math.random() - 0.5);
+let arr = [];
```
zonzujiro, line 49: Common mistake. Variable name

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+        checkMatch();
+      }, 1000);
+    }
+  }
+};
+const checkMatch = () => {
+  if (
+    !(
+      clicksStorage[0].id.replace(/\d+/g, "") ===
+      clicksStorage[1].id.replace(/\d+/g, "")
```
zonzujiro, line 75: And why you need regexp here? 🤔 I mean, you can just compare ids, without it. No?

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+    )
+  )
+    checkUnmatch();
+  clicksStorage = [];
+};
+const checkUnmatch = () => {
+  clicksStorage.forEach((target) => {
+    target.closest(".card").classList.add("card-back");
+    target.closest(".card").classList.remove("hide", "open");
```
zonzujiro, line 84: Instead of running search on in the loop, you can run it once and save result into variable

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+const createCard = ({ imgId, src }, i) => {
+  return `
+    <li class="card__container">
+          <div class="card card-back">
+              <img src="./img/${src}" id="${imgId}${i}">
+          </div>
+      </li>`;
+};
+const flipCard = (target) => {
+  if (target.closest(".card")) {
```
zonzujiro, line 59: I would move this into listener so `flipCard` will just flip cards. Also, it's a good example of unnecessary nested logic - why do you call function for cases in which it will do nothing? Better not call it all.

submissions/Eugene-CG/memory-pair-game/index.js
zonzujiro, line 46: It will be better to move into separate function. Just for readability

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
+const cardsList = document.querySelector(".cards__list");
+let clicksStorage = [];
```
zonzujiro, line 2: `const`

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+const startGame = () => {
+  let cards = [
```
zonzujiro, line 5: `const`. Variable is not changed anywhere.

Always use `const`, except situations in which you will rewrite value in it.

submissions/Eugene-CG/memory-pair-game/index.js
```diff
@@ -0,0 +1,92 @@
...
+  });
+  cardsList.innerHTML = cardListInnerHtml;
+};
+const shuffleCards = (cards) => cards.sort(() => Math.random() - 0.5);
+let arr = [];
+const createCard = ({ imgId, src }, i) => {
+  return `
+    <li class="card__container">
+          <div class="card card-back">
+              <img src="./img/${src}" id="${imgId}${i}">
```
zonzujiro, line 54: Instead of concatination, you can use `data-img-id=${imgId}`. And then compare by them. And there is no requirement that `data-*` should be unique

submissions/Eugene-CG/memory-pair-game/index.js
zonzujiro: Also, it's same as at line 61-62. So I would move this into separate function - DRY

zonzujiro (changes requested):

submissions/Eugene-CG/memory-pair-game/index.js
Eugene-CG: I cant declare clicksStorage variable with const keyword bcs I change clicksStorage later(69 line)

zonzujiro (approved):
Good job! Done

PR #141, opened 2022-08-13
uploaded files from JS DOM task

SergeyShytikov:
### JS-DOM 

[Demo](https://sergeyshytikov.github.io/js-dom-task/) |
[Code base](https://github.com/SergeyShytikov/js-dom-task)

Please, review.

### DIFF
--- a/submissions/SergeyShytikov/js-dom/app.js
+++ b/submissions/SergeyShytikov/js-dom/app.js
@@ -1,65 +1,75 @@
 const mtbCategories = [
   {
     name: "Cross-Country",
-    description: "Cross-country is perhaps the most famous discipline in mountain biking as it's the one you'll see at the Olympic Games, though there are actually several sub-categories of cross-country racing.",
-    image: "assets/cross-country.jpg"
+    description:
+      "Cross-country is perhaps the most famous discipline in mountain biking as it's the one you'll see at the Olympic Games, though there are actually several sub-categories of cross-country racing.",
+    image: "assets/cross-country.jpg",
   },
   {
     name: "Downhill",
-    description: "The name is a giveaway here; in downhill mountain bike racing, you start the track at the top of the mountain and race until you reach the bottom. Whoever completes the race course in the quickest time wins.",
-    image: "assets/downhill.jpg"
+    description:
+      "The name is a giveaway here; in downhill mountain bike racing, you start the track at the top of the mountain and race until you reach the bottom. Whoever completes the race course in the quickest time wins.",
+    image: "assets/downhill.jpg",
   },
   {
     name: "Freeride",
-    description: "Freeride makes for an incredible spectacle, with riders performing huge jumps, flips, tricks while riding on more natural terrain - such as the dramatic red rocks and cliffs of Zion National Park in Utah, USA where the Red Bull Rampage event takes place. ",
-    image: "assets/freeride.jpg"
+    description:
+      "Freeride makes for an incredible spectacle, with riders performing huge jumps, flips, tricks while riding on more natural terrain - such as the dramatic red rocks and cliffs of Zion National Park in Utah, USA where the Red Bull Rampage event takes place. ",
+    image: "assets/freeride.jpg",
   },
   {
     name: "Slopestyle",
-    description: "Closely related to freeride, slopestyle is all about demonstrating aerial tricks on a specially designed course of huge ramps, drops, jumps and berms. In fact, the clue is in the name; how much style can a rider demonstrate while heading down the slope.",
-    image: "assets/slopestyle.jpg"
+    description:
+      "Closely related to freeride, slopestyle is all about demonstrating aerial tricks on a specially designed course of huge ramps, drops, jumps and berms. In fact, the clue is in the name; how much style can a rider demonstrate while heading down the slope.",
+    image: "assets/slopestyle.jpg",
   },
   {
     name: "4X & dual slalom",
-    description: "Four-cross, usually abbreviated to 4X, is a mountain bike race discipline that takes place on a track similar to a BMX track. Four riders race simultaneously and head-to-head, with the winner the first rider to cross the finish line. In dual-slalom, two riders compete on identical parallel tracks, similar in style to 4X tracks. ",
-    image: "assets/four-cross.jpg"
+    description:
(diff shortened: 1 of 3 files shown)


submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,70 @@
...
+    description: "Is the practice of riding bikes over jumps made of dirt or soil and becoming airborne.",
+    image: "assets/dirt.jpg"
+  },
+]
+
+
+const main = document.querySelector(".main")
+const menuItems = document.querySelector(".menu-items");
+
+const createListOfLi = () => {
```
artem-trubin, line 43: This code only runs once right after you create this function. Why wrap it in function at all?

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,70 @@
...
+  })
+}
+createListOfLi()
+
+const description = document.createElement("p");
+  description.classList.add("text");
+  main.append(description);
+
+function displayImageAndText() {
+  const  arg = [...arguments].join();
```
artem-trubin, line 61: It seems you have two spaces between `const` and `arg` which is strange, set your VSCode to autoformat code for you on save.

artem-trubin, line 61: Don't use `arguments`. It is an old way of gathering arguments for functions that don't have a set amount of arguments. Use something like 
```js
function test (...myArgs) {...
  console.log(myArgs) // prints [123, 321, 231] in console
}

test(123, 321, 231)
```

artem-trubin, line 61: You don't need this here at all. You have a function that takes one thing and does something with it. Rewrite the function so it demands category name as an argument.

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,70 @@
...
+  const category = mtbCategories.find(item => item.name === arg);
+  const img = document.getElementById("image");
```
artem-trubin, line 63: You can get this `img` outside the function once, since it's always on the page, and just change its src here.

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,70 @@
...
+  img.src = category.image;
+  description.innerText = category.description;
+  return ;
```
artem-trubin, line 66: Javascript functions return `undefined` by default, no need to explicitly `return` in the end.

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,70 @@
...
+}
+createListOfLi()
+
+const description = document.createElement("p");
+  description.classList.add("text");
+  main.append(description);
+
+function displayImageAndText() {
+  const  arg = [...arguments].join();
+  const category = mtbCategories.find(item => item.name === arg);
```
artem-trubin, line 62: If possible you can write more descriptive variable names. Like here `mtbCategories` consists of `category` items. Use it instead of `item`. In that way, things like `category.name` will read as actual English and increase overall code readability.

submissions/SergeyShytikov/js-dom/index.html
```diff
@@ -0,0 +1,26 @@
...
+</head>
+<body>
+  <header class="header">
+    <h1 class="title">Mountain Biking Categories</h1>
+  </header>
+  <div class="wrapper">
+    <nav class="nav-menu">
+      <ul class="menu-items"></ul>
+    </nav>
+  <main class="main">
```
artem-trubin, line 19: In lines 19 through 22, there should be one more indentation.

submissions/SergeyShytikov/js-dom/style.css
```diff
@@ -0,0 +1,96 @@
+* {
```
artem-trubin, line 1: Add empty lines between blocks of CSS code, making the file more readable.

artem-trubin (commented):
Hey, nice job! Let's make it even better, check comments below.

submissions/SergeyShytikov/js-dom/app.js
SergeyShytikov: Thanks, I got the same comment in review about next task. I'll definitely remember that.

submissions/SergeyShytikov/js-dom/app.js
SergeyShytikov: It seems like my prettier doesn't works well.

SergeyShytikov:
Thank you a lot for such a detailed review

SergeyShytikov:
@artem-trubin Please check my PR))I understood that my design was not very inspiring, I was focused on JS

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,65 @@
...
+  const button = document.createElement("button");
+  li.classList.add("menu-item");
+  button.classList.add("button");
+  button.innerText = item.name;
+  li.appendChild(button);
+  menuItems.append(li);
+})
+
+const description = document.createElement("p");
+  description.classList.add("text");
```
artem-trubin, line 55: No need for such indentations, these should be on the same level as line 54.

submissions/SergeyShytikov/js-dom/app.js
```diff
@@ -0,0 +1,65 @@
...
+const img = document.getElementById("image");
+
+mtbCategories.forEach(item => {
+  const li = document.createElement("li");
+  const button = document.createElement("button");
+  li.classList.add("menu-item");
+  button.classList.add("button");
+  button.innerText = item.name;
+  li.appendChild(button);
+  menuItems.append(li);
```
artem-trubin, line 51: Instead of appending every item ot the page separately, we can shorten our interaction with DOM here to 1 append. You just need to create some container element, append all buttons to it and then append this container to the page in one action. You can read about [document fragment](https://developer.mozilla.org/ru/docs/Web/API/Document/createDocumentFragment), it allows you to group elements without actually creating a DOM element in the real page.

artem-trubin (changes requested):
Hey, sorry for the delay. Nice work on those changes. There are still things to improve, please check the comments below:

submissions/SergeyShytikov/js-dom/app.js
SergeyShytikov: finally I set up prettier.

artem-trubin (approved):
Good job!

PR #144, opened 2022-08-14
popup-task

filoxera1409:
### popup-task

[Demo](https://filoxera1409.github.io/hoolie-popup/) |
[Code base](https://github.com/filoxera1409/hoolie-popup)

Please, review.

### DIFF
--- a/submissions/MariiaArtamonova/html-css-popup-task/index.html
+++ b/submissions/MariiaArtamonova/html-css-popup-task/index.html
@@ -1,0 +1,126 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="./style.css" />
+    <title>Cat Popup</title>
+  </head>
+  <body>
+    <nav class="menu">
+      <ul class="nav__menu">
+          <li class="menu__item">
+            <a href="#" class="menu__item__link"><span>Mail</span></a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__item__link"><span>Images</span></a>
+          </li>
+          <li class="menu__item">
+              <input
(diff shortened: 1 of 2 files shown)


filoxera1409:
> Re-write almost everything. Re-organised and simplified the structure. Deleted unnecessary elements, classes and attributes. Added focus to non-link elements. 
Cleaner, simpler and easier, I hope.

artem-trubin (changes requested):
Hi! The demo looks completely broken right now. See screenshot below.
image

I see an error regarding the image path. Maybe it's because it has `.ru` in the address. Please store any pictures you want to use in the repo.

Fix it and request a review again.

filoxera1409:
Hi! Sorry for a delay. Added image, deleted alt descriptions. 
Thank you

artem-trubin (changes requested):
Hey, the demo still looks wrong, I assume there must be pictures in the top bar navigation, but there is an error in the console, so I suppose the images are not found. Please fix it.

Also, if we presume that pictures are fixed, I don't see any reaction of the popup button and "more" button when going through elements with the keyboard, please add one. 

About code: in CSS please add an empty line between blocks for better readability and in HTML please fix these:
- all elements on the same level in the hierarchy should have the same amount of indentations
- all indentations should have a fixed size
- all items inside other tags should have 1 more indentation than their parent

filoxera1409:
Hi! 
About image error. I saw in console some error related to "favicon.ico", but it disappeared for me, after I started to google this problem. I checked on different browsers and multiple times - and I faced 0 error in console.

So, let's clarify, what is missing for you, because for me everything is present both locally and via Git Pages.

_Web-pages include:_
_1. background image of the cat;
2. 3 icons in menu: paw, bell, and user-cat icon;
3. inside checkbox I have 9 icons of cats, more icon and 6 cats inside more;_

image

**Focus - done;
Empty lines in CSS - done;
Indentations in HTML - done;**

Thank you for your help!

artem-trubin:
I'm not sure what exactly was the problem but now it is fixed. Maybe your GitHub pages required rebuilding after committing or something. Looks great!

PR #145, opened 2022-08-14
js-dom homework by LuckyDnepr

LuckyDnepr:
### js-dom-task

[Demo](https://luckydnepr.github.io/DOM-KottansFE/) |
[Code base](https://github.com/LuckyDnepr/DOM-KottansFE)

Please, review.


submissions/LuckyDnepr/js-dom/aside-menu-action.js
```diff
@@ -0,0 +1,84 @@
...
+
+import { readJSON } from "./readJSON.js";
+
+const dataBaseUrl = "./data-films.json";
+
+export async function loadContentByID(id, container) {
+  const data = await readJSON(dataBaseUrl);
+  container.innerHTML = "";
+  const film = data.find((film) => film.ID === +id);
+  container.appendChild(makeTitle(film));
```
artem-trubin, line 11: That's a lot of reaching out to DOM just to switch content on one action. Consider wrapping all these elements with a [document fragment](https://developer.mozilla.org/ru/docs/Web/API/Document/createDocumentFragment) and using only one `appendChild`.

submissions/LuckyDnepr/js-dom/main-script.js
```diff
@@ -0,0 +1,63 @@
...
+
+async function makeAsideMenu(source, container) {
+  const data = await readJSON(source);
+  const asideMenuList = document.createElement("ul");
+  asideMenuList.classList.add("aside_menu_list");
+  data.forEach((film) => {
+    asideMenuList.appendChild(makeAsideMenuItem(film.ID, film.title));
+  });
+  container.appendChild(asideMenuList);
+  return asideMenu;
```
artem-trubin, line 24: This looks a little confusing. The purpose of this function is to render an AsideMenu. When you name function `makeAsideMenu` it usually means that it should return some kind of element, and this function does do that, but you're not using this element in any way, because what it also does is render the element on line `23` right away. Either rename the function to something like `renderAsideMenu` and remove the return or keep the return, remove rendering here and render somewhere outside this function using the returned value.

artem-trubin (changes requested):
Outstanding work! Let's improve it a little, check the comments below.

LuckyDnepr:
Thanks for review.
I solved the problems you pointed out.
Please, re-review.

artem-trubin:
Good job!

PR #149, opened 2022-08-14
html-css-popup

nazarski:
### HTML-CSS-POPUP

[Demo](https://nazarski.github.io/HTML_CSS-Popup/) |
[Code base](https://github.com/nazarski/HTML_CSS-Popup/tree/master)

Please, review.

### DIFF
--- a/submissions/nazarski/html-css-popup/style/style.css
+++ b/submissions/nazarski/html-css-popup/style/style.css
@@ -1,0 +1,184 @@
+
+/*Let's start*/
+*{
+    box-sizing: border-box;
+}
+body, html, nav, ul{
+    padding: 0;
+    margin: 0;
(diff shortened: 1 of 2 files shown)


submissions/nazarski/html-css-popup/style/style.css
```diff
@@ -0,0 +1,184 @@
+
+/*Let's start*/
```
artem-trubin, line 2: Empty line at the start and a redundant comment.

artem-trubin (changes requested):
Awesome design!

There is a problem, though. For some reason, I can select the whole popup when going through items with the keyboard. This should not be a thing, since the popup itself isn't interactive. Check the screenshot below.
image

PR #150, opened 2022-08-14
Building a Tiny JS World

LuckyDnepr:
### Building a Tiny JS World

[Demo](https://luckydnepr.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/LuckyDnepr/a-tiny-JS-world)

Please, review.


artem-trubin:
Hey, you have files from another task in this PR, please remove them.

LuckyDnepr:
@artem-trubin Is it OK now? I dont understand, how did this happen.

submissions/LuckyDnepr/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,137 @@
...
+        return makePrintableSpan("0", field);
+      }
+    }),
+    ...infoFunctionFields.map((field) => {
+      if (inhabitant.hasOwnProperty(field)) {
+        return makePrintableSpan(inhabitant[field](), field);
+      } else {
+        return makePrintableSpan("0", "field");
+      }
+    }),
```
OleksiyRudenko, line 121: Quite a volume of code repetition here. As a peer developer on the project I would love to have this fragment more lightweight.

Consider the following pattern for handlers:
```
const fieldTypeHandlers = {
  'array': list => ...,
  'function': func => ...,
  'scalar': value => ...,
}

```

OleksiyRudenko (changes requested):
@LuckyDnepr great job!

Let's make this code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

LuckyDnepr:
@OleksiyRudenko 
Maybe I should put the procedure of adding <friends> and <enemies> into a separate function and add setter and getter to these object fields?

submissions/LuckyDnepr/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,160 @@
...
+        )
+      );
+    }
+  }
+  return printableInfo.join(", ");
+}
+//assign handlers for each field type
+function fieldTypeHandler(field) {
+  let typeOfField;
+  Array.isArray(field) ? (typeOfField = "array") : (typeOfField = typeof field);
```
OleksiyRudenko, line 121: This is incorrect use of ternary operator.

Make sure `target = condition ? valueWhenConditionTrue : valueWhenConditionFalse` is used instead of `condition ? target = valueWhenConditionTrue : target = valueWhenConditionFalse`; [Conditional (ternary) operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

submissions/LuckyDnepr/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,160 @@
...
+    case "number":
+    case "boolean":
+    case "symbol":
+    case "undefined":
+      return makePrintableSpan;
+    case "object":
+      return getPrintableInhabitantInfo;
+    default:
+      break;
+  }
```
OleksiyRudenko, line 137: This can be a declarative (preferred) dictionary instead of imperative (should be avoided as much as possible) `switch`.

OleksiyRudenko (approved):
@LuckyDnepr great job!
A bit an overkill here but works.
When you re-approach this World in OOP excercize, consider following:
1) employ ES6 methods and patterns working with arrays (e.g. `[...arr, newElem]` to add elements; `.forEach/map` instead of `for...of/in`)
2) `Object` methods like `keys`, `values`, `entries` shouldn't be used when a particular order is required as these do not guarantee any particular order of keys/values. Same refers to `for...of` and `for...in` when applied to objects.
3) too much checks on whether an array is array. On data presentation stage transform data early. This helps to reduce the number of data types to process and consequently reduces the complexity of logic. [KISS](https://dev.to/kwereutosu/the-k-i-s-s-principle-in-programming-1jfg)

Also consider comment below.

submissions/LuckyDnepr/a-tiny-js-world/index.js
LuckyDnepr: Yes, I see it now.
It must be
typeOfField = Array.isArray(field) ? "array" : typeof field;

PR #152, opened 2022-08-14
HTML-CSS-practice-popup_task

Halv27:
### HTML-CSS-practice-popup_task

[Demo](https://halv27.github.io/html-css-practice-popup/) |
[Code base](https://github.com/Halv27/html-css-practice-popup)

Please, review.


Halv27:
> Hey!
> 
> Congratulations with your PR! 😎😎😎
> 
> Please, be sure you haven't followed [common mistakes](https://kottans.org/documentation/docs/doc/code-review/#popup-task).
> 
> Also, be aware, that if you would silently ignore this recommendation, a mentor may think that you are still working on fixes. And your PR will not be reviewed. 😒

Hi)) I checked the code, there are no typical mistakes in it.

artem-trubin (changes requested):
Hey, nice work! Let's improve it!

First of all, you popup button doesn't have any visual response for focusing it with keyboard.

Secondly, you cannot get to `More` button with keyboard, focus just jumps from `Documents` button to the Bell icon in the top bar.

About SVGs - your way of storing them inside the file is cool, but there could be some pros and cons about that. What's bad is that it messes up with file readability a little and those icons cannot be cached to improve performance. On the other hand, if you store them separately they might be lost somewhere since it creates more HTML requests, so you have to consider different perspectives when working with SVG when you get to real work. The common usage is to store all SVGs in one sprite and the use [use](https://developer.mozilla.org/ru/docs/Web/SVG/Element/use). (pun intended)

You can leave SVGs unchanged here, it's fine for the first work.

artem-trubin:
Good job!

Halv27:
> Good job!

Thanks))

PR #153, opened 2022-08-14
HTML & CSS Practice

kotlyar-andrey:
### HTML & CSS Practice

[Demo](https://kotlyar-andrey.github.io/html-css-popup/) |
[Code base](https://github.com/kotlyar-andrey/html-css-popup)

Please, review.

### DIFF
--- a/submissions/kotlyar-andrey/html-css-popup/style.css
+++ b/submissions/kotlyar-andrey/html-css-popup/style.css
@@ -1,0 +1,166 @@
+* {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.header {
+  background-color: #eeeeee;
+  padding: 12px;
+  display: flex;
+  list-style: none;
+  justify-content: space-between;
+  align-items: center;
+  position: relative;
+}
+.header__logo {
+  padding: 3px;
+  border-radius: 50%;
+}
+
+.header__logo:hover,
+.header__logo:focus {
+  background-color: #ffffff;
+}
+
+.header__menu {
+  display: flex;
+  list-style: none;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.header__menu > li {
+  display: block;
+  margin-right: 20px;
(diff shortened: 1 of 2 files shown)


kotlyar-andrey:
Hi!
I only noticed the absence of nested lists.
Please let me know if there is anything else

submissions/kotlyar-andrey/html-css-popup/style.css
```diff
@@ -0,0 +1,167 @@
...
+
+.popup_more {
+  display: none;
+}
+
+#more_checkbox:checked ~ .more_label,
+#more_checkbox:checked {
+  display: none;
+}
+#more_checkbox:checked ~ .popup_more {
```
artem-trubin, line 160: It is fine here, but for the future please stick to the concept of dividing responsibilities: 
- _classes_ are for _CSS_ and _styling_
- _ids_ are for _JS_ and _functionality_
- _tags_ are for _HTML_ and _layout_

submissions/kotlyar-andrey/html-css-popup/style.css
```diff
@@ -0,0 +1,167 @@
...
+  display: block;
+}
+
+hr.more_hr {
```
artem-trubin, line 164: I believe only `hr`s would have `.more_hr` class, so why not just `.more_hr` here?

artem-trubin (changes requested):
Hi, Nice work! Let's improve a little, you should be consistent about your coding styling. Make sure your CSS block are separated by empty lines for better readability. Also, check the comments below.

kotlyar-andrey:
Thanks for review!
I fixed it, re-review please.

submissions/kotlyar-andrey/html-css-popup/style.css
kotlyar-andrey: id is also used here for the label's attribute "for", so I really left it. Thanks for the advice!

artem-trubin:
Good job!

PR #154, opened 2022-08-14
Add dom-task

HelenGreent:
### js-dom

[Demo](https://helengreent.github.io/kottans-js-dom/) |
[Code base](https://github.com/HelenGreent/kottans-js-dom.git)

### DIFF
--- a/submissions/HelenGreent/js-dom/index.js
+++ b/submissions/HelenGreent/js-dom/index.js
@@ -1,0 +1,73 @@
+const herbs = [
+  {
+    name: "Aconitum",
+    blossoming: "June-August",
+    location: "North America, Europe, and Asia",
+    image: "https://media.istockphoto.com/photos/wildflowers-picture-id846069016?k=20&m=846069016&s=612x612&w=0&h=dN4M66iDCb8sT-eZq-k4omrBjno0jzvBKMLKwYh0s9g=",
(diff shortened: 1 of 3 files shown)


submissions/HelenGreent/js-dom/index.js
```diff
@@ -0,0 +1,73 @@
...
+        <img class='image' src='${item.image}' alt='${item.name}'/>
+        <figcaption class='figcaption'>${item.name}</figcaption>
+    </figure>
+    </div>
+    <p class='description'>${item.desc}</p>`;
+  return article;
+}
+
+nav.addEventListener("click", (event) => {
+  let currentLink;
```
artem-trubin, line 65: What is happening here? `currentLink` will always be `undefined` and the condition below will always work, since `currentLink` is created each time you click a button and immediately gets deleted since it is not used outside this function. Maybe you wanted to prevent the user from reclicking the same button, but it didn't work. You should move the `currentLink` outside this function so it can store a value. Also, if you want something to be `empty`, use `null` value. `undefined` usually means someone forgot to place a value or something is lost.

artem-trubin (changes requested):
Hi, nice work. Let's make it even better! Please read the comments below.

- There is a strange behavior with menu buttons: when you hover over the whole button, a border appears, and then when you hover over the text it lights up. It's a little unintuitive from UX perspective since it makes the user think there are two separate meanings to the clicks on the text and on the whole button. Maybe make text light up when the whole button is hovered over, or remove the text highlighting at all. 
- There is no way to tell which menu option is chosen right now. Maybe after you fix your `currentLink` problem specified in code, you can fix this one.

artem-trubin:
Nice changes! Good job.

PR #159, opened 2022-08-15
Document Object Model - practice

zhenyakornilov:
### Document Object Model - practice

[Demo](https://zhenyakornilov.github.io/kottans-js-dom-practice/) |
[Code base](https://github.com/zhenyakornilov/kottans-js-dom-practice)

Please, review.


zhenyakornilov:
Removed DOM calls from loop in renderButtons function

zhenyakornilov:
Could you give some advice how to load image faster?

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,422 @@
...
+  padding: 0;
+}
+
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+ul,
```
kasionio, line 12: this style applies just for `li` tags, so you can leave just one tag here
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,422 @@
...
+  color: inherit;
+  text-decoration: none;
+  outline: none;
+}
+
+button {
+  background: none;
+  color: inherit;
+  border: none;
+  padding: 0;
```
kasionio, line 32: you specify this property for all elements at the first lines

submissions/zhenyakornilov/js-dom/index.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE html>
+<html lang="en" id="root">
```
kasionio, line 2: It is a common practice to divide responsibility - ids for JS, classes for styling. Better to use the wrapper in this case and set styles for `wrapper`, not for `html`tag. Using html styling is even bad for performance reasons

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,422 @@
...
+}
+
+.menu-items {
+  margin-top: 50px;
+}
+
+.nav-item {
+  border-bottom: 1px solid #f5f5dc;
+  width: 20rem;
+  margin: 10;
```
kasionio, line 180: what are the units?

submissions/zhenyakornilov/js-dom/index.html
```diff
@@ -0,0 +1,47 @@
...
+  <link rel="icon" type="image/x-icon" href="images/museum-icon.svg">
+  <link rel="stylesheet" href="css/styles.css">
+  <title>Top museums of Kyiv</title>
+</head>
+<body class="body">
+  <div class="wrapper">
+    <header class="header">
+      <span class="dropdown-wrap">
+        <button class="dropdown">
+          <i class="fa-solid fa-bars fa-lg"></i>
```
kasionio, line 20: fa-solid fa-bars fa-lg. What are these classes for?

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,422 @@
...
+.content-body .desc-wrapper {
+  grid-area: desc;
+  line-height: 1.6;
+  font-size: 1.75rem;
+  overflow-wrap: break-word;
+  margin-top: 20px;
+}
+
+.content-body .desc-wrapper .museum-image {
+  float: left;
```
kasionio, line 248: `float` is an old stuff and nowadays better to avoid using it, you can use flex instead

submissions/zhenyakornilov/js-dom/index.html
```diff
@@ -0,0 +1,47 @@
...
+        <input class="check-switch" type="checkbox" id="switcher">
+        <label class="theme-switcher" for="switcher">
+          <i class="fa-solid fa-sun"></i>
+          <i class="fa-solid fa-moon"></i>
+          <span class="circle"></span>
+        </label>
+      </div>
+    </header>
+    <nav class="sidebar">
+      <ul class="menu-items"></ul>
```
kasionio, line 34: you can add `id` here and use it in JS `getElementById`

submissions/zhenyakornilov/js-dom/js/data.js
```diff
@@ -0,0 +1,82 @@
...
+    description:
+      "The National Museum “Chornobyl” is a multifunctional institution combining scientific, cultural and educational " +
+      "activity with a modern museum and archive, documenting,preservingand conveying the history of the Chornobyl " +
+      "nuclear disaster as the most severe radioecological disaster of the XX century. The after-effects of it have " +
+      "no analogy and differ from other natural or manmade catastrophes. These are still the issues of contemporaneity " +
+      "and, unfortunately, of the generation to come." +
+      "The Museum opened to the public on April 26, 1992,in Kiev, the capital of Ukraine, from a hundred " +
+      "kilometers of the epicenter of the disaster.",
+    imageSrc: "images/chernobyl.jpg",
+    museumUrl: "http://chornobylmuseum.kiev.ua/uk/",
```
kasionio, line 14: just `url` will be better

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
+"use strict";
+
+import museums from "./data.js";
+import switchTheme from "./utils.js";
```
kasionio, line 4: I don't see such function in utils

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+function main() {
+  switchTheme();
+
+  const listMenu = document.querySelector(".menu-items");
```
kasionio, line 9: please use id for this purpose. id is unique identifier

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
...
+        selectedButton = this;
+
+        renderArticle(getContextObj(), content);
+      }
+    });
+  }
+}
+
+function renderArticle(museumObj, content) {
+  const contentHTML = `
```
kasionio, line 63: You can use object destructuring 
 https://javascript.info/destructuring-assignment#object-destructuring

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
...
+      `<li class="nav-item">` +
+      `  <button class="btn-text">${btnName}</button>` +
+      `</li>`;
+  });
+
+  menu.innerHTML = buttonsHTML;
+}
+
+function toggleSelectedButton(museums, content) {
+  let selectedButton = document.getElementsByClassName("nav-item")[1];
```
kasionio, line 32: Logically, default value is the first element of the array, not second

submissions/zhenyakornilov/js-dom/js/utils.js
```diff
@@ -0,0 +1,12 @@
+function changeTheme() {
+  
+  const checkbox = document.getElementById("switcher");
+  checkbox.addEventListener("change", () => {
```
kasionio, line 4: better to separate the handler function from a listener. 
el.addEventListener("change", handler);
function handler {}

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
...
+  const listMenu = document.querySelector(".menu-items");
+  const contentBody = document.querySelector(".content-body");
+  renderButtons(museums, listMenu);
+  toggleSelectedButton(museums, contentBody);
+  openMenu(listMenu);
+}
+
+document.addEventListener("DOMContentLoaded", main);
+
+function renderButtons(arr, menu) {
```
kasionio, line 18: arr sounds too generic. We can name it `listOfCards` for example or what you prefer more

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,88 @@
...
+  let buttonsHTML = ``;
+
+  arr.forEach(({ btnName }) => {
```
kasionio, line 21: `map` looks like the optimal choice

kasionio (changes requested):
Nice job, looking good! Please take a look at the comments. And I found some weird behavior. A highlighted zone is also active but doesn't have a hover effect
image

kasionio:
About images - the simplest way is to compress images or use a lightweight format such as webp. One more option - while an image is loading you can set some default placeholder instead. Also, HTML has a cool tag `picture`, take a look on it
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture

submissions/zhenyakornilov/js-dom/index.html
zhenyakornilov: can I use :root pseudo-class here then?

submissions/zhenyakornilov/js-dom/index.html
zhenyakornilov: It's Font Awesome prebuilt classes, I thought it is not needed to change them.

submissions/zhenyakornilov/js-dom/css/styles.css
zhenyakornilov: I used float here to wrap text around image, I haven't figured out how to do that with flex yet.

zhenyakornilov:
Tried to make all fixes accordingly to remarks.

zhenyakornilov:
I have questions about some remarks, see in comments above.

zhenyakornilov:
> Nice job, looking good! Please take a look at the comments. And I found some weird behavior. A highlighted zone is also active but doesn't have a hover effect image

this one looks to be fixed too

submissions/zhenyakornilov/js-dom/index.html
kasionio: the common use of :root is for variables declaration. For such files root will be always `html`
https://developer.mozilla.org/en-US/docs/Web/CSS/:root

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,89 @@
...
+      ({ btnName }) => btnName === selectedBtnText.textContent
+    )[0];
+
+    return obj;
+  }
+
+  renderArticle(getContextObj(), content);
+
+  for (let button of buttons) {
+    button.addEventListener("click", function () {
```
kasionio, line 50: here you added a listener to every button. why not add a listener to the navigation menu?

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,89 @@
...
+    let obj = museums.filter(
+      ({ btnName }) => btnName === selectedBtnText.textContent
+    )[0];
+
+    return obj;
+  }
+
+  renderArticle(getContextObj(), content);
+
+  for (let button of buttons) {
```
kasionio, line 49: try to use a different approach. `for...of` is not needed here. now it seems that you take different parts of code from different sources

kasionio (changes requested):
Great progress, but let's improve JS part 🚀

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,89 @@
...
+  menu.innerHTML = buttonsHTML;
+}
+
+function toggleSelectedButton(museums, content) {
+  let selectedButton = document.getElementsByClassName("nav-item")[0];
+  selectedButton.classList.add("open");
+
+  const buttons = document.getElementsByClassName("nav-item");
+
+  function getContextObj() {
```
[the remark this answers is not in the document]
zhenyakornilov: This function returns object from museums array to render as content.

submissions/zhenyakornilov/js-dom/index.html
kasionio: you are right. my bad

submissions/zhenyakornilov/js-dom/css/styles.css
kasionio: Let's leave it, it is not a big deal. Try to use this tag here, it is a common way of using it, add a caption

submissions/zhenyakornilov/js-dom/js/index.js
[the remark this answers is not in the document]
kasionio: I mean name. getConteNtObject?

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,420 @@
...
+    width: 450px;
+  }
+}
+
+@media screen and (max-width: 720px) {
+  .wrapper {
+    grid-template-areas: "header-top header-top" "content-main content-main" "footer-bottom footer-bottom";
+  }
+
+  .dropdown {
```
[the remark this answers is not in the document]
zhenyakornilov: Sorry, I don't understand this remark here.

Do You mean something wrong with **grid-template-areas**?

submissions/zhenyakornilov/js-dom/css/styles.css
```diff
@@ -0,0 +1,420 @@
...
+button {
+  background: none;
+  color: inherit;
+  border: none;
+  font: inherit;
+  cursor: pointer;
+  outline: inherit;
+}
+
+:root {
```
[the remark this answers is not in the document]
zhenyakornilov: I change html font-size since I use **rem** and when html font-size is equal to 62.5% then 1 rem is like 10 pixels, 2.1rem is 21 pixels etc. It makes more clear to me rem values to interpret them as pixels.

submissions/zhenyakornilov/js-dom/css/styles.css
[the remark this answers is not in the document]
kasionio: You have two similar classes

submissions/zhenyakornilov/js-dom/css/styles.css
zhenyakornilov: Did You mean to place museum description to **figcaption** tag here instead of **\<p\>** tag?

submissions/zhenyakornilov/js-dom/css/styles.css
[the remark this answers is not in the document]
zhenyakornilov: Yet I don't get what I have to change here(

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,89 @@
+"use strict";
```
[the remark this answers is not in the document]
zhenyakornilov: Changed index.js. Did You mean something like it looks now?

zhenyakornilov:
Tried to make changes accordingly to recommendations, but still have some questions above.

kasionio:
Please do not resolve comments, I need to see them

submissions/zhenyakornilov/js-dom/js/index.js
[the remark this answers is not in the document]
kasionio: Not yet. Divide listeners and functions 
element.addEventListener(event, **listener**);
Look below, the **listener** is bold 

<pre>function showMuseumInfoByClick(museums, content, menu) {
  menu.addEventListener("click",<b>function ({ target }) {
    if (target.tagName.toLowerCase() === "button") {
      const currentContentObject = museums.filter(
        ({ btnName }) => btnName === target.textContent
      );
      renderArticle(currentContentObject[0], content);
    }
  }</b>);
} </pre>

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,68 @@
...
+      renderArticle(currentContentObject[0], content);
+    }
+  });
+}
+
+function renderArticle({ name, description, imageSrc, url }, content) {
+  const contentHTML = `
+  <h2 class="museum-title">${name}</h2>
+  <div class="desc-wrapper">
+    <img class="museum-image" src="${imageSrc}" alt="photo of museum">
```
kasionio, line 46: `alt` should be moved to `data.js` as one of the paramters, because it can't be the same for different images

submissions/zhenyakornilov/js-dom/js/index.js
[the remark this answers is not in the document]
zhenyakornilov: separated listeners and functions

submissions/zhenyakornilov/js-dom/css/styles.css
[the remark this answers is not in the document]
zhenyakornilov: I don't understand what's wrong here. I defined **.dropdown** in CSS with display none to hide and then changed this in media query to make it visible.

submissions/zhenyakornilov/js-dom/index.html
```diff
@@ -0,0 +1,49 @@
...
+          <i class="fa-solid fa-moon moon-icon"></i>
+          <span class="circle"></span>
+        </label>
+      </div>
+    </header>
+    <nav class="sidebar">
+      <ul class="menu-items" id="menu"></ul>
+    </nav>
+    <main class="content">
+      <article class="content-body" id="content"></article>
```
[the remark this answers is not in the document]
zhenyakornilov: I have misunderstanding here. I have **h2** tag added with JS inside **article** tag, isn't it can be counted as header? Or shall I set h2 tag inside **index.html**? May I just  change **article** tag to div then?

zhenyakornilov:
New chnages added, but some questions left.

submissions/zhenyakornilov/js-dom/index.html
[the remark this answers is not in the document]
kasionio: you are right, I missed `h2`. All is fine here

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,67 @@
...
+      `;
+    })
+    .join("");
+
+  menu.innerHTML = buttonsHTML;
+}
+
+function showMuseumInfo({ target }) {
+  if (target.tagName.toLowerCase() === "button") {
+    const currentContentObject = museums.filter(
```
kasionio, line 27: you can try to use spread here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,67 @@
...
+      </li>
+      `;
+    })
+    .join("");
+
+  menu.innerHTML = buttonsHTML;
+}
+
+function showMuseumInfo({ target }) {
+  if (target.tagName.toLowerCase() === "button") {
```
kasionio, line 26: why do not use `.closest` here?

kasionio: also better to prevent rendering when you click on the same button which already selected

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,67 @@
...
+  sidebar.classList.toggle("active");
+}
+
+function main() {
+  renderButtons(museums, listMenu);
+  renderArticle(museums[0], contentBody);
+
+  themeSwitcher.addEventListener("change", toggleDarkClass);
+  dropdownBtn.addEventListener("click", openDropdownMenu);
+  listMenu.addEventListener("click", showMuseumInfo);
```
kasionio, line 64: good article to understand the events bubbling process 
https://javascript.info/bubbling-and-capturing

kasionio (changes requested):
Much better now, great job!
Some small changes left. And about the hover effect on switchTheme button. All interactive elements should have `pointer` cursor on hover

submissions/zhenyakornilov/js-dom/js/index.js
zhenyakornilov: Thanks

zhenyakornilov:
Added latest changes

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,77 @@
...
+import museums from "./data.js";
+import toggleDarkClass from "./utils.js";
+
+const themeSwitcher = document.getElementById("switcher");
+const listMenu = document.getElementById("menu");
+const dropdownBtn = document.getElementById("dropdown-button");
+const contentBody = document.getElementById("content");
+
+function renderButtons(museumsObj, menu) {
+  let buttonsHTML = museumsObj
```
kasionio, line 12: if you have plural, better name it like buttonsListHtml

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,77 @@
...
+      `;
+    })
+    .join("");
+
+  menu.innerHTML = buttonsHTML;
+}
+
+function showMuseumInfo({ target }) {
+  if (target.closest(".btn-text")) {
+    let currentSelectedBtn = target.closest(".nav-item");
```
kasionio, line 27: you can move it higher and then make instead of two one `if` statement

submissions/zhenyakornilov/js-dom/js/index.js
```diff
@@ -0,0 +1,77 @@
...
+    if (!currentSelectedBtn.classList.contains("active")) {
+      const [currentContentObject] = museums.filter(
+        ({ btnName }) => btnName === target.textContent
```
kasionio, line 30: Better use `id` to identify the target item, not the text. For example, we can have two items with the same button name. So, in such cases we should always use id for DOM manipulation

kasionio (changes requested):
Let's make last small improvements ;)

zhenyakornilov:
Added new changes accordingly to remarks.

kasionio (approved):
Fair enough, I guess :) Good job 👍

PR #161, opened 2022-08-15
Tiny JS World

natalia-artsiukh:
### Tiny JS World

[Demo](https://natalia-artsiukh.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/natalia-artsiukh/a-tiny-JS-world)

Please, review.


submissions/nataliaArtsiukh/TinyJSWorld/lib.js
```diff
@@ -0,0 +1,68 @@
...
+  var targetDomain = (domainComponents.length > 1)
+    ? 'github.com/' + domainComponents[0]
+    : domainComponents[0];
+  return prefix + '//' + targetDomain + basePath + (domainComponents.length > 1 ? 'blob/gh-pages/' : '') + filePath;
+}
+
+(function (elementId) {
+  var element = document.getElementById(elementId);
+  element.href = makeGitHubLink(location.href, 'index.js');
+})('source-code');
```
OleksiyRudenko, line 67: I wonder why would you need to duplicate these functions?
No harm tho, not an issue in this case.

submissions/nataliaArtsiukh/TinyJSWorld/lib.js
```diff
@@ -0,0 +1,68 @@
...
+  hands: 2,
+  legs: 2,
+  name: 'Halle',
+  gender: 'female',
+  friends: 'Mary, Tom, Batman'
+};
+
+const habitants = [man, woman, cat, dog, catwoman];
+
+habitants.forEach(element => {print(element.saying + '! My name is ' + element.name + '. I\'m ' + element.gender + '. I have ' + element.hands + ' hands and ' + element.legs + ' legs. I\'m friends with ' + element.friends + '.'); return;});
```
OleksiyRudenko, line 44: Please rewrite this fragment with [JS template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

Also consider [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) in arrow function arguments to make the code fragment less verbose.

Also you do not need an "empty" `return` statement. By default, functions return `undefined` if no `return` statement provided (which is actually equivalent to having `return;`).

OleksiyRudenko (changes requested):
@natalia-artsiukh a good start!
A couple of fixes and we are done.

submissions/nataliaArtsiukh/TinyJSWorld/lib.js
natalia-artsiukh: These functions were in this file originally. I didn't dare to remove them.

submissions/nataliaArtsiukh/TinyJSWorld/lib.js
natalia-artsiukh: And i don't understand how to destructure an arrow function if it takes only one argument.

submissions/nataliaArtsiukh/TinyJSWorld/lib.js
OleksiyRudenko: Oh, I see now. Didn't expect to see `lib.js` changed...

image

Please (a) move your code to `index.js`, (b) restore `lib.js` to its prior state, (c) add `index.js` to this PR and (d) remove `lib.js` from this PR.

submissions/nataliaArtsiukh/TinyJSWorld/lib.js
OleksiyRudenko: Consider the following examples (you will still need an effort to build an arrow function based on that).
```
const obj = {
  prop1: 'abc',
  prop2: 123,
}; // "structuring" usage of object definition syntax

const { prop1, prop2 } = obj; // left-hand part is destructuring due to its context (assignment)
console.log(prop1, prop2); // Console output: 'abc', 123

function doImportantAction(o) {
  console.log(o.prop1, o.prop2); // traditional reference to properties
}
doImportantAction(obj); // Console output: 'abc', 123

function doAnotherImportantThing({ prop1, prop2 }) { // parameters list is destructuring due to its context (params list)
  console.log(prop1, prop2);
}
doImportantAction(obj); // Console output: 'abc', 123
```

So, what you do with a function is you destructure an parameter of Object type it expects as an input.

natalia-artsiukh: Thanks for explanation! Everything is clear now)

submissions/nataliaArtsiukh/TinyJSWorld/index.js
```diff
@@ -0,0 +1,44 @@
...
+ const dog = {
+   hands: 0,
+   legs: 4,
+   name: 'Rex',
+   gender: 'male',
+   saying: 'Woof',
+   friends: 'Mary'
+ };
+ const catwoman = {
+   __proto__:cat,
```
OleksiyRudenko, line 34: When you go through JS OO module (prototypical inheritance) please pay special attention to how correctly bind prototypes.

You do not need to fix anything here at this stage.

OleksiyRudenko (approved):
@natalia-artsiukh great job!

PR #163, opened 2022-08-15
DOM homework

bmukha:
[Demo](https://bmukha.github.io/kottans-dom/) |
[Code base](https://github.com/bmukha/kottans-dom)

Please, review.


submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,81 @@
...
+  },
+];
+
+const renderHero = (hero) => {
+  const heroClass = hero.class.toLowerCase().split(' ').join('');
+  return `<h2 class="hero-name">${hero.name}</h2>
+    <div class="hero-portrait">
+        <img class="hero-portrait_img" src="./img/${heroClass}.webp" alt="heroportrait" />
```
madmaxWMFU, line 58: Read and try [dectruction](https://javascript.info/destructuring-assignment)

submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+    <p class="lore">
+        ${hero.lore}
+    </p>`;
+};
+
+const content = document.querySelector('.content');
+
+document.addEventListener('DOMContentLoaded', () => {
+  const randomNumber = Math.floor(Math.random() * 9);
```
madmaxWMFU, line 71: A magic number. Use a `const` of array length.

submissions/bmukha/dom/index.html
```diff
@@ -0,0 +1,86 @@
...
+    <title>Kottans DOM practice</title>
+    <link rel="stylesheet" href="./css/style.css" />
+  </head>
+  <body class="main-wrapper">
+    <header class="header">Heroes of Hearthstone</header>
+    <aside class="menu">
+      <nav class="nav">
+        <ul class="list">
+          <li class="list-item">
+            <div id="demonhunter" class="hero-info">
```
madmaxWMFU, line 16: It is a link. Maybe you can use the tag `<a></a>`

submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,81 @@
...
+  content.innerHTML = renderHero(heroes[randomNumber]);
+});
+
+document.querySelector('.menu').addEventListener('click', (event) => {
+  const hero = event.target.parentElement.id;
+  const info = heroes.find(
+    (item) => item.class.toLowerCase().split(' ').join('') === hero
```
madmaxWMFU, line 78: In your list of heroes one hero with class consists of two words. Maybe you rename and write less code? It only for discussion. How do you think?

madmaxWMFU (changes requested):
Good work. See some comments below.

bmukha:
First of all, thx for the feedback, I appreciate it a lot. Regarding Demon Hunter, the only hero with two words in a class name. Well, he was very problematic to deal with from the beginning. I've tried to simplify that class name in the code everywhere I could, but it should definitely stay intact on the main page. People will try to find me if I misspell one of the Word of Warcraft classes intentionally, you know ) Another possible solution I can think of from the top of my head is to use something like data- in my HTML as hero's class name container, but I don't think it worth the effort to make this change just to get rid of a couple of lines of code. All the other issues that you've pointed out, have been addressed. Please, re-review.

submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,92 @@
...
+
+document.addEventListener('DOMContentLoaded', () => {
+  const randomNumber = Math.floor(Math.random() * heroes.length);
+  content.innerHTML = renderHero(heroes[randomNumber]);
+});
+
+document.querySelector('.menu').addEventListener('click', (event) => {
+  const hero = event.target.parentElement.id;
+  const info = heroes.find(
+    (item) => item.heroClass.toLowerCase().split(' ').join('') === hero
```
madmaxWMFU, line 89: The variable `info` hasn't correct information.
`const infoHero = heroes.find(heroe => {})`

submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,92 @@
...
+
+const content = document.querySelector('.content');
+
+document.addEventListener('DOMContentLoaded', () => {
+  const randomNumber = Math.floor(Math.random() * heroes.length);
+  content.innerHTML = renderHero(heroes[randomNumber]);
+});
+
+document.querySelector('.menu').addEventListener('click', (event) => {
+  const hero = event.target.parentElement.id;
```
madmaxWMFU, line 87: You read and changed the code used destruction. In this part, you also can try this. 
`document.querySelector('.menu').addEventListener('click', ({target}) => {}`

madmaxWMFU (changes requested):

submissions/bmukha/dom/app.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const content = document.querySelector('.content');
+
+document.addEventListener('DOMContentLoaded', () => {
+  const randomNumber = Math.floor(Math.random() * heroes.length);
+  content.innerHTML = renderHero(heroes[randomNumber]);
+});
+
+document.querySelector('.menu').addEventListener('click', (event) => {
```
madmaxWMFU, line 86: Maybe is a better to set the listener on element 'list'

madmaxWMFU:
Correct me if I'm slightly wrong.
First of all, a dataset is a good option for saving some information that we may need to work with.
Second, you already have a static menu with the names of the heroes. That is, in your array, the `heroClass` property would be better called `heroId`. The user would not be able to see it, it is unlikely that he would crawl into the source code. When you check, you compare the id attribute of the element with the array property 
```javascript
const hero = event.target.parentElement.id;  // demonhunter
...
//item.heroClass.toLowerCase().split(' ').join('')
// Demon Hunter -> demonhunter -> ['demon', hunter] -> demonhunter
item.heroClass.toLowerCase().split(' ').join('') === hero // demonhunter == demonhunter
...

```

bmukha:
Tnx again for the suggestions. It kinda stuck in my head from the beginning of the project, that I use all those hero class names in the dynamic part of the page. Apparently, I am not. Of course, it makes perfect sense to just rename stuff in my array and get rid of all the string manipulation code altogether. Shame on me for not noticing this. So, I've changed the code to address this and all the other suggestions. Please, take a look.

madmaxWMFU (approved):
Great, I approve of your work. Good luck with future tasks!
```
const getContent = ({ target }) => {
  content.innerHTML = renderHero(
    heroes.find((item) => item.heroId === target.parentElement.id)
  );
}
document.querySelector('.list').addEventListener('click', );

const init = () => {
  const randomNumber = Math.floor(Math.random() * heroes.length);
  content.innerHTML = renderHero(heroes[randomNumber]);
}

document.addEventListener('DOMContentLoaded', init);
```

PR #190, opened 2022-08-16
tiny js world 

SergeyShytikov:
### tiny JS world 

[Demo](https://sergeyshytikov.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/SergeyShytikov/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/SergeyShytikov/tiny-js-world-task/index.js
+++ b/submissions/SergeyShytikov/tiny-js-world-task/index.js
@@ -70,5 +70,7 @@ const keysOrder = ['species', 'name', 'gender', 'legs', 'hands', 'saying'];
    print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
    */
 
-const mappedInhabitants = inhabitants.map(obj => keysOrder.map(key => obj[key])) ;
-mappedInhabitants.forEach(item => print(item.join('; ')));
+   inhabitants
+   .map(inhabitant => keysOrder
+      .map(key => inhabitant[key]))
+   .forEach(arrayOfValues => print(arrayOfValues.join('; ')));
--- a/submissions/SergeyShytikov/tiny-js-world-task/index.js
+++ b/submissions/SergeyShytikov/tiny-js-world-task/index.js
@@ -1,0 +1,68 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+const dog = {
+   species: 'dog',
+   name: 'Ice',
+   gender: 'male',
+   legs: 4,
+   hands: 0,
+   saying: 'woof-woof!'
+};
+const cat = {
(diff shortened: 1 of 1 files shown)


submissions/SergeyShytikov/tiny-js-world-task/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+
+function getPropertiesValueToString(obj) {
+return print(Object.values(obj).join('; '))
+}
```
OleksiyRudenko, line 66: Looks like a generic method to convert any object into a string.

However, as the task requires specific order of properties listed `Object` methods like `keys`, `values`, `entries` shouldn't be used as these do not guarantee any particular order of keys/values.

Once you solve this the method will stop being generic but rather specific to our domain (Tiny JS World and its inhabitants). Thus you will need to reflect this specificity in function and its parameter names.

OleksiyRudenko (changes requested):
@SergeyShytikov good job!
Please check the comment below.

submissions/SergeyShytikov/tiny-js-world-task/index.js
SergeyShytikov: Thanks for the advice, for some reason I did not read all the requirements of the task.

submissions/SergeyShytikov/tiny-js-world-task/index.js
```diff
@@ -0,0 +1,74 @@
...
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+
+const mappedInhabitants = inhabitants.map(obj => keysOrder.map(key => obj[key])) ;
+mappedInhabitants.forEach(item => print(item.join('; ')));
```
OleksiyRudenko, line 74: If we chain `inhabitants.map` and following `forEach` we do not need an extra variable.
`obj` is too generic. As a peer developer I want to **read** instantly what objects we work here with.
Same is about `item`.

Do not use generic variable names unless it is a function specifically designed to work with multiple types of instances (e.g. we are building a generic data processing library).

OleksiyRudenko (changes requested):
@SergeyShytikov well done!
A few things to improve and we are done.

submissions/SergeyShytikov/tiny-js-world-task/index.js
SergeyShytikov: You right, after a while I won't be able to figure out what's going on here, either.

OleksiyRudenko (approved):
@SergeyShytikov well done, grats!

P.S. Please install prettier.io plugin on your IDE for it to take care of formatting and code style for you. On bigger codebase it is worth it definitely.

PR #204, opened 2022-08-17
Frogger-game

zhenyakornilov:
### Object-Oriented JS - Frogger-game

[Demo](https://zhenyakornilov.github.io/kottans-frogger-game-oop/) |
[Code base](https://github.com/zhenyakornilov/kottans-frogger-game-oop)

Please, review.

### DIFF
--- a/submissions/zhenyakornilov/frogger-game/app.js
+++ b/submissions/zhenyakornilov/frogger-game/app.js
@@ -28,6 +28,7 @@ class Character {
     ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
   }
 }
+
 class Enemy extends Character {
   constructor(x, y, speed, direction, sprite) {
     super(x, y, sprite);
@@ -38,7 +39,6 @@ class Enemy extends Character {
   update(dt) {
     if (this.direction === "left") {
       this.x -= dt * this.speed;
-
       if (this.x < 0) {
         this.x = GAME_BOARD_WIDTH + 10;
       }
@@ -59,7 +59,6 @@ class Enemy extends Character {
       player.y < this.y + 60 &&
       player.y + 60 > this.y
     ) {
-      console.log(player.x);
       player.x = PLAYER_START_POSITION.x;
       player.y = PLAYER_START_POSITION.y;
     }
--- a/submissions/zhenyakornilov/frogger-game/app.js
+++ b/submissions/zhenyakornilov/frogger-game/app.js
@@ -16,6 +16,8 @@ const MOVE_STEP = {
 };
 
 const GAME_BOARD_WIDTH = 505;
+const OUTBOARD_INDENT = 10;
+const SAFE_DISTANCE = 60;
 
 class Character {
   constructor(x, y, sprite) {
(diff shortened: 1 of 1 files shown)


zhenyakornilov:
### Covered some common mistakes:
1. Class **Enemy** do not refer to global **player** variable, **player** passed to **Enemy** constructor instead.
2. **Enemy** **update()** method do not call **checkOverlap()** method inside. Moved **checkOverlap()** method to **updateEntities** function in **engine.js** instead.

OleksiyRudenko:
> 2\. **Enemy** **update()** method do not call **checkOverlap()** method inside. Moved **checkOverlap()** method to **updateEntities** function in **engine.js** instead.

Please bring it back to the `app.js` for the sake of code review scope. It can perfectly be a part of Enemy class (as it is specific to Enemy behaviour), just needs to be a method on its own.

submissions/zhenyakornilov/frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+    this.speed = speed;
+    this.direction = direction;
+    this.player = player;
+  }
+
+  update(dt) {
+    if (this.direction === "left") {
+      this.x -= dt * this.speed;
+      if (this.x < 0) {
+        this.x = GAME_BOARD_WIDTH + 10;
```
OleksiyRudenko, line 44: WHat is `10` here and across the codebase? I guess `-10` is the same but just negative. If true then having `OFFSET = 10` and referring to it as `OFFSET` and `-OFFSET` would be fine.
Of course `OFFSET` word doesn't explain it application, so invent a better name.

submissions/zhenyakornilov/frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+      this.x += dt * this.speed;
+      if (this.x > GAME_BOARD_WIDTH) {
+        this.x = -10;
+      }
+    }
+  }
+
+  checkOverlap() {
+    if (
+      this.player.x < this.x + 60 &&
```
OleksiyRudenko, line 56: What is `60`?

OleksiyRudenko (changes requested):
@zhenyakornilov very well coded!

Apart of bringing collision checking function back as mentioned above, let's add more semantics to the constants so that peer developers were able to **read** what every number means and implement changes quickly when needed.

zhenyakornilov:
Made some changes

OleksiyRudenko (changes requested):
@zhenyakornilov please go through your code and extrapolate previous comments re magic numbers to all other similar situations in your code.
Do not expect a reviewer to "red pen" every single tiny bug in the code. Own it truly.

zhenyakornilov:
It seems to be a bit better now.

OleksiyRudenko (approved):
@zhenyakornilov a really good job!

Hint on naming in graphic environments:
In regular structures like matrices, game fields (which are also matrices) elements' coordinates are defined by row (or column), matrix row height (or matrix column width), and offset.
So, for e.g. `y = verticalOffset + row * rowHeight`.  Items to place into the matrix may have their own individual offsets (like in this game where visual part has offsets against its pixel representation).
So having defined offsets, widths and heights it is possible to calculate any coordinates based on row and column index.

PR #208, opened 2022-08-18
tiny-js-world

Vlada-khom:
### Tiny-js-World

[Demo](https://vlada-khom.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Vlada-khom/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Vlada-khom/tiny-js-world/index.js
+++ b/submissions/Vlada-khom/tiny-js-world/index.js
@@ -1,0 +1,78 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+let man = {
+  entity: "person",
+  name: "John",
+  sex: "male",
+  hands: 2,
+  legs: 2,
+  say: "Hello world!",
+  friends: "Jenny, Rex",
+};
+
+let woman = {
+  entity: "person",
+  name: "Jenny",
+  sex: "female",
+  hands: 2,
+  legs: 2,
+  say: "Hello world! I'm Jenny.",
+  friends: "Tom",
+};
+
+let dog = {
+  entity: "dog",
+  name: "Rex",
+  sex: "male",
+  hands: 0,
+  legs: 4,
+  say: "Gav!",
+  friends: "John, Jenny",
+};
+
+let cat = {
+  entity: "cat",
+  name: "Tom",
+  sex: "male",
+  hands: 0,
+  legs: 4,
+  say: "Meyow!",
+  friends: "Jenny",
+};
+
+// ======== OUTPUT ========
+/* Use print(message) for output.
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+function makeString(obj) {
+  let arrStr = [];
+  for (let value in obj) {
+    arrStr.push(obj[value]);
+  }
+  return arrStr.join(";");
+}
+
+print(makeString(man));
+print(makeString(woman));
+print(makeString(cat));
+print(makeString(dog));
+/* Print examples:
+   print('ABC');
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
--- a/submissions/Vlada-khom/tiny-js-world/index.js
+++ b/submissions/Vlada-khom/tiny-js-world/index.js
@@ -55,9 +55,11 @@ let cat = {
    so code reviewers might focus on a single file that is index.js.
    */
 
-const arrObjects = [man, woman, cat, dog];
-let list = arrObjects.map((el)=> [el.entity, el.name, el.sex, el.hands, el.legs, el.say, el.friends].join('; '));
-list.forEach(el=> print(el));
+   const inhabitants = [man, woman, cat, dog];
+   const properties = ['entity', 'name', 'sex', 'hands', 'legs', 'say', 'friends'];
+   let characteristicsOfInhabitants = inhabitants.forEach((inhabitant)=> {
+     print(properties.map(property => [inhabitant[property]]).join('; '))
+     });
 
 /* Print examples:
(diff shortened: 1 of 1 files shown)


submissions/Vlada-khom/tiny-js-world/index.js
```diff
@@ -0,0 +1,78 @@
...
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+function makeString(obj) {
+  let arrStr = [];
+  for (let value in obj) {
+    arrStr.push(obj[value]);
+  }
+  return arrStr.join(";");
```
OleksiyRudenko, line 63: `Object` methods like `keys`, `values`, `entries` shouldn't be used when a particular order is required as these do not guarantee any particular order of keys/values. Same refers to `for...of` and `for...in` when applied to objects.

submissions/Vlada-khom/tiny-js-world/index.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+print(makeString(man));
+print(makeString(woman));
+print(makeString(cat));
+print(makeString(dog));
```
OleksiyRudenko, line 69: Code should be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), which means that whenever you see a pattern in your code those should be eliminated as much as possible. 

Imagine, we have 100s of inhabitants.

Examples:
    - `print(dog); print(cat); etc ... ` should be refactored employing `Array.forEach` as the least
    - `` `${obj.legs}; ${obj.name}; etc...` `` (yes, strings are also code) must be refactored employing appropriate `Array` methods

OleksiyRudenko (changes requested):
@Vlada-khom good job!
A couple of improvements and we're done.

submissions/Vlada-khom/tiny-js-world/index.js
```diff
@@ -0,0 +1,70 @@
...
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+const arrObjects = [man, woman, cat, dog];
+let list = arrObjects.map((el)=> [el.entity, el.name, el.sex, el.hands, el.legs, el.say, el.friends].join('; '));
+list.forEach(el=> print(el));
```
OleksiyRudenko, line 60: 1. Code is written for other people. The closer the code to English the better. Give variables and functions meaningful names. Avoid generic names like `item`, `element`, `key`, `object`, `array` or their variations. Exception: helper functions that are specifically and intentionally designed to be multipurpose (not the case here).
2. `[el.entity, el.name, el.sex, el.hands, el.legs, el.say, el.friends]` is not DRY, contains repetition. As a peer developer I am tasked to add 20 more properties. At inhabitants presentation part of code I want to just list new properties without much repetition.

OleksiyRudenko (changes requested):
@Vlada-khom well done.
Let's polish it and practice "speaking" variables naming and [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

submissions/Vlada-khom/tiny-js-world/index.js
```diff
@@ -0,0 +1,72 @@
...
+   Default tag for message is <pre>. Use print(message,'div') to change containing element tag.
+
+   Message can contain HTML markup. You may also tweak index.html and/or styles.css.
+   However, please, REFRAIN from improving visuals at least until your code is reviewed
+   so code reviewers might focus on a single file that is index.js.
+   */
+
+   const inhabitants = [man, woman, cat, dog];
+   const properties = ['entity', 'name', 'sex', 'hands', 'legs', 'say', 'friends'];
+   let characteristicsOfInhabitants = inhabitants.forEach((inhabitant)=> {
```
OleksiyRudenko, line 60: What `characteristicsOfInhabitants` holds? Why it holds what it holds? And why do we need this variable?

submissions/Vlada-khom/tiny-js-world/index.js
```diff
@@ -0,0 +1,72 @@
...
+     print(properties.map(property => [inhabitant[property]]).join('; '))
```
OleksiyRudenko, line 61: What is the result of this `.map`? Observe in console
Why do we need to wrap `inhabitant[property]` in square brackets?

submissions/Vlada-khom/tiny-js-world/index.js
Vlada-khom: You are right. I can write it without this variable

submissions/Vlada-khom/tiny-js-world/index.js
Vlada-khom: Clear, I wrap up every property in array. Result map its array [person, John, male, 2, 2 and etc],[person, Jenny,...]. forEach iterate array but return nothing... at the end we must get [],[],[],[].join(transform to string)...
okay, I have questions, if methods Object.values, keys and entries doesn't guarantee ordered. Does Object.fromEntries again not guarantee ordered? If yes, when better use these methods?

OleksiyRudenko: > I wrap up every property in array
... which we do not actually need. `['entity', 'name']` should turn into `['human', 'John']`, not `[['human'], ['John']]`. Luckily, `[['a']].join(';')` turns into `'a'` due to implicit array extrapolation.
Check what is the result of `[['human', 'John'], ['male', 2, ["a","b"]]].join('; ')` to see the effect of nested arrays. At times this may be what you actually want, but often results in bugs and may trick the developer when they try to explore/debug data structures.

Guess, now you know enough to polish the code.

> if methods Object.values, keys and entries doesn't guarantee order. Does Object.fromEntries again not guarantee order? If yes, when better use these methods?

It doesn't either. They are useful when the order doesn't matter.
A use case: you need a list of some object properties that are also a part of some reference list.
```
const obj = { propF: 11,  propX: 1, propB: 7, propA: 121 };
const requiredProps = ['propA', 'propD', 'propX'].sort();
const intersectionOfPropNames = Object.keys(obj).filter(key => requiredProps.includes(key)); // ['propX', 'propA']
// next we could probably tell if all required props are present
intersectionOfPropNames.length === requiredProps.length // naive
intersectionOfPropNames.sort().join(';') === requiredProps.join(';') // naive 2
JSON.from(intersectionOfPropNames.sort()) === JSON.from(requiredProps) // advanced
requiredProps.reduce((noMisses, requiredProp) => noMisses && intersectionOfPropNames.includes(requiredProp), true) // pro
```

Use case: Transform all values in an object where properties are not predefined.
```
const envVariables = {
  apiKey: 'API_Key',
  ghActor: 'Git_User',
};

function objectValuesToUppercas(obj) {
  return Object.fromEntries(
    Object.entries(obj).map(([key, value]) => [key, value.toUpperCase()])
  );
}

objectValuesToUppercas(envVariables);
```

Vlada-khom: Clear, thank you) What is mean JSON.from in first example, I didn't find?

OleksiyRudenko: I meant `JSON.stringify`

PR #218, opened 2022-08-18
TJW part 2 ( Js post oop )

the-darkesttt:
### Js post oop

[Demo](https://the-darkesttt.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/the-darkesttt/a-tiny-JS-world)

Please, review.


submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,71 @@
...
+const inhabitants = [
+   new Dog('Robert', 'male', 'BARK!'),
+   new Cat('Lily', 'female', 'meooow!'),
+   new Human('Charlie', 'female', 'Hi John!'),
+   new Human('John', 'male', 'Hi Charlie!'),
+];
+
+inhabitants.forEach((obj) => {
+   obj.showValues(obj);
+})
```
OleksiyRudenko, line 71: Do not use generic names like `obj`, `item`. Use domain specific naming instead.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,71 @@
...
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Creature {
+   species;
+   name;
+   gender;
+   saying;
+   legs;
+   hands;
```
OleksiyRudenko, line 17: Not every creature has hands.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+   constructor( name, gender, saying) {
+      this.name = name;
+      this.gender = gender;
+      this.saying = saying;
+   }
+
+   showValues(obj) {
+      print(Object.values(obj).join('; '));
```
OleksiyRudenko, line 26: `Object` methods like `keys`, `values`, `entries` shouldn't be used when a particular order is required as these do not guarantee any particular order of keys/values. Same refers to `for...of` and `for...in` when applied to objects.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,71 @@
...
+   }
+}
+
+class Animal extends Creature {
+   constructor(name, gender, saying) {
+      super(name, gender, saying);
+      this.legs = 4;
+      this.hands = 0;
```
OleksiyRudenko, line 34: A class cannot initialize props that do not belong to it immediately.

OleksiyRudenko (changes requested):
@the-darkesttt nice try.
Please reiterate through OOP and SOLID principles [here](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) and leverage the power of OOP

OleksiyRudenko:
I started the review, but...
Output doesn't match the requirements. Demo is requested for purpose. EIther legs or hands are missing.
`Creature` has prop `saying` but it is never initialized by `Creature`.
Code is not DRY. Initialization via `super` is not consistent with `Creature`'s constructor specification.
The bot will add a checklist. Please go through it carefully.

the-darkesttt:
@OleksiyRudenko I updated PR, please review

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,92 @@
...
+      const valueKeys = [
+         'species',
+         'name',
+         'gender',
+         'saying'
+      ];
+      const inhabitantValues = valueKeys.map((key) => {
+         return `${this[key]}`;
+      });
+      return inhabitantValues;
```
OleksiyRudenko, line 29: `keys` is a technical term. Inhabitants are characterized by rather properties or features. No one says 'Name is my key'.

Also you can chain up all these statements in a single `return` statement.

If an arrow function has a single argument parentheses are not required.
Arrow functions that have a single return statement do not need curly braces and `return` keyword.
Compare:
```
names.map((name) => { return name.toUpperCase(); })

names.map(name => name.toUpperCase())
```

Why `this.key` is wrapped into the template string?

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+   showValues() {
+      const valueKeys = [
+         'paws'
+      ];
+      let inhabitantValues = valueKeys.map((key) => {
+         return `${this[key]}`;
+      });
+      let inhabitantValuesNew = super.showValues().concat(inhabitantValues);
+      return print(inhabitantValuesNew.join('; '));
```
OleksiyRudenko, line 47: What if a subclass will want its own `showValues`? Adding `print` at this point will disable further extension.
`return print(...)` doesn't make sense as `print` returns `undefined`.
Avoid side effects. Or put them aside until very end (to the calling code).

Also this method does two things - (1) build values for representation and (2) actual output. Separation of concerns is a thing.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,92 @@
...
+   showValues() {
+      const valueKeys = [
+         'legs',
+         'hands',
+      ];
+      let inhabitantValues = valueKeys.map((key) => {
+         return `${this[key]}`;
+      });
+      let inhabitantValuesNew = super.showValues().concat(inhabitantValues);
+      return print(inhabitantValuesNew.join('; '));
```
OleksiyRudenko, line 79: Same as above.
Once you find ways to optimize `Creature.showValues`, you will also be able to optimize the overloaded versions of the method in child classes.

OleksiyRudenko (changes requested):
@the-darkesttt good job!
Although a few of improvements required.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+class Animal extends Creature {
+   constructor(name, gender, saying, paws, species) {
+      super(name, gender, saying, species);
+      this.paws = paws;
+   }
+
+   showValues() {
+      return `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.paws}`
```
OleksiyRudenko, line 27: Make this DRY suing `Array#map`.
Also, class should handle its own properties. Encapsulation from OOP and Open/Closed from SOLID.

submissions/the-darkesttt/js-post-oop/index.js
```diff
@@ -0,0 +1,64 @@
...
+class Human extends Creature {
+   constructor(name, gender, saying) {
+      super(name, gender, saying, 'homosapien');
+      this.hands = 2;
+      this.legs = 2;
+   }
+
+   showValues() {
+      return `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.legs}; ${this.hands}`
+   }
```
OleksiyRudenko, line 52: See how extended `super` API can help you make this fragment DRYer.
You will find at least 3 helpful materials on `super`  [here](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up)

OleksiyRudenko (changes requested):
@the-darkesttt need to make code OOP/SOLID compliant.
Also show you've mastered `Array` methods.

submissions/the-darkesttt/js-post-oop/index.js
OleksiyRudenko: Explain, why ignored.
Imagine, you are on an important commercial project.

PR #222, opened 2022-08-19
OOP exercise

kotlyar-andrey:
### OOP exercise

[Demo](https://kotlyar-andrey.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/kotlyar-andrey/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/kotlyar-andrey/a-tiny-JS-world/index.js
+++ b/submissions/kotlyar-andrey/a-tiny-JS-world/index.js
@@ -5,72 +5,132 @@
    Web app: https://kotlyar-andrey.github.io/a-tiny-JS-world/
    */
 
-class Creature {
-  constructor(species, name, gender, legs, hands, phrase, linkTo = null) {
+/**
+ * Общий класс для всех жителей
+ */
+ class Creature {
+  constructor(species, name, gender, phrase) {
     this._species = species;
     this._name = name;
     this._gender = gender;
-    this._legs = legs;
-    this._hands = hands;
     this._phrase = phrase;
     this._friends = [];
-    this._link = linkTo;
-    this._properties = [
-      "species",
-      "_name",
-      "_gender",
-      "_legs",
-      "_hands",
-      "saying",
-      "friends",
-    ];
+    this._properties = ["_species", "_name", "_gender"];
   }
 
-  get species() {
-    return this._link
-      ? `${this._species}-${this._link._species}`
-      : this._species;
-  }
-
-  get friends() {
+  getFriends() {
     return this._friends && this._friends.length > 0
       ? this._friends.map(friend => friend._name).join(", ")
       : "no friends";
   }
 
-  set friends(newFriends) {
+  setFriends(newFriends) {
     this._friends = [...newFriends];
   }
 
-  get saying() {
-    return this._link ? this._link._phrase : this._phrase;
+  getSaying() {
+    return this._phrase;
   }
 
-  set saying(newFrase) {
-    this._phrase = newFrase;
+  setSaying(newPhrase) {
+    this._phrase = newPhrase;
   }
 
-  display() {
-    print(this._properties.map(prop => this[prop]).join("; "));
+  addProperties(...newProperties) {
+    this._properties = [...this._properties, ...newProperties];
+  }
+
+  getPropertiesData() {
+    return this._properties.map(prop => this[prop]).join("; ");
+  }
+
+  getAnotherData() {
+    const friends = this.getFriends();
+    const saying = this.getSaying();
+    return `; ${friends}; ${saying}`;
+  }
+
+  getInfo() {
+    return this.getPropertiesData() + this.getAnotherData();
   }
 }
 
-const cat = new Creature("cat", "Jerry", "male", 4, 0, "Myau");
-const dog = new Creature("dog", "Bars", "male", 4, 0, "Woof");
-const woman = new Creature("human", "Maria", "female", 2, 2, "Hi");
-const man = new Creature("human", "Ivan", "male", 2, 2, "Hello");
-const woman_cat = new Creature("human", "Marta", "female", 2, 2, "", cat);
-const werewolf = new Creature("human", "Petr", "male", 4, 2, "", dog);
+/**
+ * Общий класс для всех животных с ногами/лапами
+ */
+class Animal extends Creature {
+  constructor(species, name, gender, phrase, legs) {
+    super(species, name, gender, phrase);
+    this._legs = legs;
+    this.addProperties("_legs");
+  }
+}
 
-man.friends = [woman, cat, dog];
-woman.friends = [man, cat];
-dog.friends = [cat];
-woman_cat.friends = [man, cat];
+/**
+ * Общий класс для птиц
+ */
+class Bird extends Animal {
+  constructor(species, name, gender, phrase) {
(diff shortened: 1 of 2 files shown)


submissions/kotlyar-andrey/oop_exercise/index.js
```diff
@@ -0,0 +1,106 @@
...
+      this._properties
+        .map((prop) => this[prop])
+        .join("; ")
+        .concat(`; ${friends}`)
+    );
+  }
+}
+
+class HomoSapiens extends Creature {
+  constructor(species, name, gender, phrase, legs, hands) {
```
OleksiyRudenko, line 52: Do we really want users to specify species for class Homo Sapiens?

submissions/kotlyar-andrey/oop_exercise/index.js
```diff
@@ -0,0 +1,106 @@
...
+    this._hands = hands;
+    this.addProperties("_legs", "_hands");
+  }
+
+  get saying() {
+    return this._phrase;
+  }
+}
+
+class Animal extends Creature {
```
OleksiyRudenko, line 64: Strictly speaking humans are animals as well. We need to resolve the matter in a consistent way.

OleksiyRudenko (changes requested):
@kotlyar-andrey very well done!
OOP is not really good for building hybrids. And this is one of the learnings from this exercise to take. The proper approach would be composition. You do not need to fix anything in these regards. Just reflect on possible flaws of OOP-like approach.

Still we need to add some consistency with regards to the business domain. Please check comments below.

kotlyar-andrey:
Thanks for the review!
I've made changes, please re-review.

submissions/kotlyar-andrey/oop_exercise/index.js
```diff
@@ -0,0 +1,104 @@
...
+        .concat(`; ${friends}`)
+    );
+  }
+}
+
+class Animal extends Creature {
+  constructor(species, name, gender, phrase, legs, hands) {
+    super(species, name, gender, phrase);
+    this._legs = legs;
+    this._hands = hands;
```
OleksiyRudenko, line 55: Do all animals have hands?
Imagine we popultae the world with birds, fish, rhinos and deer.
OOP helps to model the real world quite accurately avoiding defining properties and behavior that the counterparts in the real world do not have.

OleksiyRudenko (changes requested):
@kotlyar-andrey good fix. Yet we need to make this code compliant with OOP/SOLID principles.
You will find extra materials [here](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) helpful.
Feel free asking specific questions based on your learnings.

kotlyar-andrey:
I think, I understand what you mean. Review, please)

submissions/kotlyar-andrey/oop_exercise/index.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  addProperties(...newProperties) {
+    this._properties = [...this._properties, ...newProperties];
+  }
+
+  display() {
+    const friends = this.getFriends();
+    const saying = this.getSaying();
+    print(
```
OleksiyRudenko, line 46: A class shouldn't know anything about external environment.
Using a side effect (`print` is a side effect) also makes the class less flexible.

submissions/kotlyar-andrey/oop_exercise/index.js
```diff
@@ -0,0 +1,133 @@
...
+      this._properties
+        .map(prop => this[prop])
+        .join("; ")
+        .concat(`; ${saying}; ${friends}`)
```
OleksiyRudenko, line 50: Not a really good approach to have a list of non-array props and handle array props in a totally different way. Besides, this method does two things: converts array props to strings, builds object presentation string.

What if we have 20 more array props? What if we have props that keep objects?

We need a more consistent approach.

Suggest an approach before implementing changes in these regards.

OleksiyRudenko (changes requested):
@kotlyar-andrey very good.
Although there are two flaws that will make the project harder to maintain and scale if we want to expand it.

submissions/kotlyar-andrey/oop_exercise/index.js
kotlyar-andrey: Ok, I can rename display to getInfo and return value
  ```
getInfo() {
    const friends = this.getFriends();
    const saying = this.getSaying();
    return this._properties
      .map(prop => this[prop])
      .join("; ")
      .concat(`; ${saying}; ${friends}`);
  }
```
  End then print inhabitants
 ```
 inhabitants.forEach(creature => {
    print(creature.getInfo());
});
```

OleksiyRudenko: Making class method just return some data is definitely a good design decision.

submissions/kotlyar-andrey/oop_exercise/index.js
kotlyar-andrey: > Not a really good approach to have a list of non-array props and handle array props in a totally different way. Besides, this method does two three things: converts array props to strings, builds object presentation string, prints the presentation string.

We can refactor this method to three another function

> What if we have 20 more array props? What if we have props that keep objects?

Every class has addProperties method for adding new properties. 
If we want to get another data in child classes, we can override method getAnotherData

 ```
getDataProperties() {
    return this._properties.map(prop => this[prop]).join("; ");
  }

  getAnotherData() {
    const friends = this.getFriends();
    const saying = this.getSaying();
    return `; ${friends}; ${saying}`;
  }

  getInfo() {
    return this.getDataProperties() + this.getAnotherData();
  }
```

This is all what I can offer

File: submissions/kotlyar-andrey/a-tiny-JS-world/index.js [hunk 6]
```diff
@@ -5,72 +5,132 @@
...
+
+/**
+ * Общий класс для всех животных с ногами/лапами
+ */
+class Animal extends Creature {
+  constructor(species, name, gender, phrase, legs) {
+    super(species, name, gender, phrase);
+    this._legs = legs;
+    this.addProperties("_legs");
+  }
```
OleksiyRudenko, line 66: If some class (e.g. `Creature`) is extended by one and only one class (e.g. `Animal`) this means they can be effectively merged. Yes, we may have legless creatures in future, but that's exactly the point when we would split a base class and/or add yet another base class on the top of hierarchy tree.

submissions/kotlyar-andrey/a-tiny-JS-world/index.js
```diff
@@ -5,72 +5,132 @@
...
+    this._wings = 2;
+    this.addProperties("_wings");
+  }
+}
+
+class Human extends Animal {
+  constructor(name, gender, phrase) {
+    super("human", name, gender, phrase, 2);
+    this._hands = 2;
+    this.addProperties("_hands");
```
OleksiyRudenko, line 84: We delegate child class property handling to a base class.
This is a legal approach. Although we should always consider first that every class handles its own and only its own properties. This way the handling logic will be encapsulated within the owning class and not spread across the hierarchy.

OleksiyRudenko (approved):
@kotlyar-andrey great job done!
Takes notes of the comments below.

PR #224, opened 2022-08-19
character and enemies created

natalia-artsiukh:
### Object Oriented JS

[Demo](https://natalia-artsiukh.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/natalia-artsiukh/frontend-nanodegree-arcade-game)

To make the collection of stars to complete the game, I made changes to engine.js. Two lines have been added there to update and render star objects similar to enemy objects. It was possible to implement this within app.js but this would break the structure of the entire code.

Please, review.


submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+    };
+    this.handleInput = function(direction) {
+        if (this.y > 0) {
+            if (direction == 'left' && this.x > 0) {
+                this.x -= colWidth;
+            } else if (direction == 'up' && this.y > 0) {
+                this.y -= rowHeight;
+            } else if (direction == 'right' && this.x < colWidth * 4) {
+                this.x += colWidth;
+            } else if (direction == 'down' && this.y < rowHeight * 5) {
```
OleksiyRudenko, line 99: What are `4` and `5`?

submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+var player;
+
+function launch() {
+    allStars = [];
+    for (var i = 0; i < 5; i++) {
+        allStars.push(new Star(i));
+    }
+    
+    allEnemies = [];
+    for (var i = 1; i <= 3; i++){
```
OleksiyRudenko, line 121: What are `5` and `3`?

submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+
+// Now write your own player class
+// This class requires an update(), render() and
+// a handleInput() method.
+var Player = function() {
+    this.x = colWidth * 2;
+    this.y = rowHeight * 5 - 10;
+    this.finished = false;
+    this.sprite = 'images/char-boy.png';
+    this.update = function(){
```
OleksiyRudenko, line 68: In JS prototypical class model methods are properly declared in a different way.
Check [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain). Feel free to find other materials that would explain the model better.

BTW, how happened methods for Star and Enemy are defined with proper syntax?

submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+        for (var i = 0; i < allEnemies.length; i++) {
+            if (this.y == allEnemies[i].y + 10) {
+                if (allEnemies[i].x > this.x - colWidth + 20 && allEnemies[i].x < this.x + colWidth - 20) {
+                    this.x = colWidth * 2;
+                    this.y = rowHeight * 5 - 10;
+                } 
+            }
+        }
```
OleksiyRudenko, line 76: (1) Use `Array` methods whenever possible, in general.
(2) Probably it makes more chance that it is an enemy checks if they collided with the Player, rather than Player scans all enemies.
(3) Classes should not refer to any global variables, like global variable `player` or `enemies`, which is an instance of `Player` class or contains instances of `Enemy` class.

(referring to global constants and globals provided by the gaming platform like `Resources` is OK);

Hint: you can pass `Player` instance as an argument to every enemy.

(4) Separation of Concerns principle should be followed.  (e.g. `update` method does only rendering and doesn't contain any unrelated **inline** code; for example collision check is defined as a dedicated method and only called from inside `update`)

submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,141 @@
...
+    };
+}
+
+// Now instantiate your objects.
+// Place all enemy objects in an array called allEnemies
+// Place the player object in a variable called player
+
+var allStars;
+var allEnemies;
+var player;
```
OleksiyRudenko, line 112: Switch to `const` and `let` syntax.

OleksiyRudenko (changes requested):
@natalia-artsiukh very well done!
Yet a couple of things to improve.

OleksiyRudenko:
On a personal note, I like that your implementation of gameplay goes beyond the minimal requirements.

submissions/nataliaArtsiukh/Frogger-game/app.js
natalia-artsiukh: I wrote in the style of the original code, but now code is corrected)

submissions/nataliaArtsiukh/Frogger-game/app.js
natalia-artsiukh: I assumed that since the player exists in a single instance, the prototype is not needed. Now everything is fixed.

submissions/nataliaArtsiukh/Frogger-game/engine.js
```diff
@@ -0,0 +1,202 @@
...
+        });
+        player.update();
+    }
+
+    function checkCollisions() {
+        if (allEnemies.some(enemy => enemy.isCollided(player))) {
+            player.comeBack();
+        };
+    }
+
```
OleksiyRudenko, line 109: We do not want to change the game's engine. Please restore the code and make collision testing methods for enemies and stars part of respective classes.

It is OK to call collision test method from another (e.g. `update`) method. The point is to separate concerns/function.

submissions/nataliaArtsiukh/Frogger-game/app.js
```diff
@@ -0,0 +1,168 @@
...
+
+const Star = function(numCol) {
+    this.x = sizes.colWidth * numCol; 
+    this.y = -sizes.starVerticalShift;
+    this.sprite = 'images/Star.png';
+}
+
+Star.prototype.update = function() {
+    if (player.y < 0) {
+        let index = player.x / sizes.colWidth;
```
OleksiyRudenko, line 25: Here the code implies global `player` variable exists.
Classes shouldn't know anything outer world unless they get this information explicitly (e.g. a reference to a foreign object is passed to them at construction time and stored among class properties for re-use).

Direct reference to global environmental variables (like `Resources`) is OK here. This is a flaw of the game engine architecture, but we do not want to fix it. This happens on real-world projects.

OleksiyRudenko (changes requested):
@natalia-artsiukh now the code reads lamost in plain English, which is very good.

A couple of OOP/SOLID related fixes and we are done.

submissions/nataliaArtsiukh/Frogger-game/app.js
OleksiyRudenko: The comment above remains valid. Key part is "at construction".
Engine's `update` function has its own signature. It doesn't pass any variables to game characters and we do not want to interfere with it.
image

The only way to resolve is to pass the `player` at object instantiation time, store it in enemies/stars instances properties just like coordinates or whatever we store locally and re-use it (local reference to the player) in update/collision testing method.

natalia-artsiukh: I made the necessary changes. I can't click re-review. please check)

OleksiyRudenko (approved):
@natalia-artsiukh perfectly done!

OleksiyRudenko:
On a side note: commit messages are thoughtful and informative. Very professional habit and skill

PR #229, opened 2022-08-19
wrong merging path

vovan-zt:
### popup

[Demo](https://vovan-zt.github.io/popup2022/) |
[Code base](https://github.com/vovan-zt/popup2022.git)

wrong merging path


artem-trubin:
Don't just add new files, remove old ones.

artem-trubin:
For passing-by mentors, this PR is for moving files that are already merged, just with a wrong path.

OleksiyRudenko:
@vovan-zt normally renaming with `git mv` does the job

vovan-zt:
then I need to do it on the first branch that was confirmed and delete this branch?

OleksiyRudenko:
You must have the wrong-named files on your `main` where you branched off to create this PR.
If it is true then just delete those, commit and update this PR.
If it is not true then you'd need to update your `main`, rebase this branch on top of it and make the situation true ;)
This way we emulate files renaming with `git mv`.

OleksiyRudenko:
Now it looks all good.
image

Other changes are minor (formatting), although a few empty lines look redundant. But I guess we can live with it.

@artem-trubin WDYT?

PR #232, opened 2022-08-20
add frogger

bmukha:
### frogger

[Demo](https://bmukha.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/bmukha/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/bmukha/frogger/app.js
+++ b/submissions/bmukha/frogger/app.js
@@ -1,51 +1,158 @@
-let highScore = localStorage.getItem('highScore') ?? 0;
-let currentScore = 0;
-const highHolder = document.querySelector('.high');
-const currentHolder = document.querySelector('.current');
-const infoHolder = document.querySelector('.info');
+class Environment {
+  constructor() {
+    this.tileWidth = 101;
+    this.tileHeight = 83;
+    this.tilesInRow = 5;
+    this.fieldMinX = 0;
+    this.fieldMinY = -10;
+    this.fieldMaxX = this.tileWidth * this.tilesInRow;
+    this.fieldMaxY = 415;
+    this.isGameOn = true;
+    this.directions = ['left', 'right'];
+  }
+  getRandomNumber(min, max) {
+    return Math.floor(Math.random() * (max - min) + min);
+  }
 
-const getRandomNumber = (min, max) =>
-  Math.floor(Math.random() * (max - min) + min);
+  getRandomDirection() {
+    return this.directions[Math.floor(Math.random() * this.directions.length)];
+  }
+}
 
-const directions = ['left', 'right'];
+class ScoreBoard {
+  constructor(highScoreHolder, messageHolder, currentScoreHolder) {
+    this.currentScore = 0;
+    this.highScoreHolder = document.querySelector(highScoreHolder);
+    this.messageHolder = document.querySelector(messageHolder);
+    this.welcomeMessage = "Let's go!";
+    this.winMessage = 'You win!';
+    this.loseMessage = 'You lose!';
+    this.highScore = localStorage.getItem('highScore') ?? 0;
+    this.currentScoreHolder = document.querySelector(currentScoreHolder);
+  }
+  updateGameInfo(message) {
+    this.currentScoreHolder.innerText = `Current score: ${this.currentScore}`;
+    this.highScoreHolder.innerText = `High score: ${this.highScore}`;
+    this.messageHolder.innerText = `${message}`;
+  }
+  incrementScore() {
+    this.currentScore += 1;
+    if (this.currentScore > this.highScore) {
+      this.highScore = this.currentScore;
+      this.saveHighScore(this.highScore);
+    }
+  }
+  saveHighScore(score) {
+    localStorage.setItem('highScore', score);
+  }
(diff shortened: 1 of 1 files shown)


submissions/bmukha/frogger/app.js
```diff
@@ -0,0 +1,149 @@
...
+
+canvas = {
+  width: 505,
+  height: 606,
+};
+
+class Enemy {
+  constructor(startY, mult) {
+    this.width = 101;
+    this.height = 171;
```
OleksiyRudenko, line 23: I think these character width and height are true not only for enemy.
Also 101 is somehow relates to `505` above. Can we reduce number of constants and build some others using calculations?
As a peer developer tasked to amend this world (number of rows, columns, dimensions of tiles/characters) i'd like to change as few number as only possible.

P.S. Check magic numbers here and across entire codebase.

submissions/bmukha/frogger/app.js
```diff
@@ -0,0 +1,149 @@
...
+    this.currX = this.startX;
+    this.currY = this.startY;
+    this.mult = this.direction === 'right' ? mult : -mult;
+  }
+  update(dt) {
+    if (
+      player.currX >= this.currX &&
+      player.currX <= this.currX + this.width &&
+      player.currY >= this.currY &&
+      player.currY <= this.currY + 85
```
OleksiyRudenko, line 39: (1) classes should not not refer to any global variables, like global variable `player`, which is an instance of `Player` class

(referring to global constants and globals provided by the gaming platform like `Resources` is OK);

Hint: pass `Player` instance as an argument to every enemy

(2) Separation of Concerns principle should be followed

(e.g. `update` method does only rendering and doesn't contain any unrelated **inline** code; for example collision check is defined as a dedicated method and only called from inside `update`)

submissions/bmukha/frogger/app.js
```diff
@@ -0,0 +1,149 @@
...
+    ) {
+      infoHolder.innerText = `You lost`;
+      currentScore = 0;
+      currentHolder.innerText = `Current score: ${currentScore}`;
```
OleksiyRudenko, line 43: Class shouldn't know anything about environment it operates in (whether it web or node engine).

OleksiyRudenko (changes requested):
@bmukha very well done!
Let's improve code semantics and make it more OOP/SOLID principles compliant.

OleksiyRudenko:
Oh, sometimes may hero gets killed right after stepping on the rock tile with no enemies close to him.
This effect starts after the first failure.
You may want to look into this bug.

bmukha:
Tnx for the comments/suggestions I really appreciate it! OOP is totally new territory for me, so, probably, implementing them will take a couple of days. A lot of reading/research needs to be done.

bmukha:
Hi. So, I made some changes to address the issues you pointed out. I believe I got rid of the respawning bug and improved the collision detection algorithm but it's probably the only things am sure about in all of this. Regarding reducing the number of constants, I succeeded only partially. The width of the field 505 can be derived from tile width 101 multiplied by the number of tiles in a row with no problem. But I failed to find such an apparent relationship on the Y axis. It looks like the real height of the png files is 171, but the step, which makes the game look nice, is 83, and the real height of the canvas is 606, which looks like 6 * 101. It's extremely confusing, I'm not gonna lie. And finally, I made collision detection a separate method and redone a couple more things into classes, but I'm far from sure that all this is correct. Somehow code looks even less clean and more bloated to me. Please, take a look.

submissions/bmukha/frogger/app.js
```diff
@@ -1,149 +1,192 @@
...
+const firstEnemyY = env.fieldMinY + env.tileHeight;
+const secondEnemyY = env.fieldMinY + env.tileHeight * 2;
+const thirdEnemyY = env.fieldMinY + env.tileHeight * 3;
 const allEnemies = [
-  new Enemy(65, getRandomNumber(150, 300)),
-  new Enemy(150, getRandomNumber(150, 300)),
-  new Enemy(235, getRandomNumber(150, 300)),
+  new Enemy(firstEnemyY, player, env, scoreBoard),
+  new Enemy(secondEnemyY, player, env, scoreBoard),
+  new Enemy(thirdEnemyY, player, env, scoreBoard),
```
OleksiyRudenko, line 182: What if I need to add 20 more enemies?
Do I need to define `fourthEnemyY` etc or this `Y`'s are actually y-coords for the field rows?

OleksiyRudenko (approved):
@bmukha great job!
Check comments below for consideration.

submissions/bmukha/frogger/app.js
```diff
@@ -1,149 +1,192 @@
...
-const infoHolder = document.querySelector('.info');
-
-const getRandomNumber = (min, max) =>
-  Math.floor(Math.random() * (max - min) + min);
-
-const directions = ['left', 'right'];
-
-const getRandomDirection = () =>
-  directions[Math.floor(Math.random() * directions.length)];
+class Environment {
```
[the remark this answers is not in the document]
bmukha: Thank you for your feedback, really appreciate it!

PR #235, opened 2022-08-20
DOM_task 

Tedzury:
### DOM_task

[Demo](https://tedzury.github.io/DOM_kottans_task/) |
[Code base](https://github.com/Tedzury/DOM_kottans_task)

Please, review.

### DIFF
--- a/submissions/Oleksii_Drohachov/DOM_task/script.js
+++ b/submissions/Oleksii_Drohachov/DOM_task/script.js
@@ -1,0 +1,162 @@
+const burger = document.querySelector('.burger');
+const menu = document.querySelector('.menu');
+const menuList = document.querySelector('.menu__list_hidden');
+const contentBox = document.querySelector('.main__content');
+
+
+const greetings = {
+    imgUrl: 'img/Lorgar.jpg',
+    text: '&emsp;Hello, my little heretic! <br><br> &ensp;I\'m here to introduce you to the fascinating world of chaos, pure chaos, where everyone can whoever ot whatever he wants. God Emperor failed us, he lied to us, but I WILL TELL YOU THE TRUTH! Sit down and make yourself comfortable. <br><br> &ensp;Where should I begin my story?'
+};
+
+const content = [
+    {name: 'Tzeentch',
+     imgUrl: 'img/Tzeentch.jpg',
+     altName: 'Lord of Change',
+     warCry: '"Do not ask which creature screams in the night. Do not question who waits for you in the shadow. It is my cry that wakes you in the night, and my body that crouches in the shadow. I am Tzeentch and you are the puppet that dances to my tune..."',
+     descr: '&emsp;Tzeentch, also known as the "Changer of Ways," the "Lord of Change," "Lord of Sorcery," and the "Architect of Fate," among many other names and titles, is the Chaos God of change, evolution, mutation, intrigue, ambition, knowledge, sorcery, destiny, lies and trickery. <br><br>&emsp;Tzeentch is especially empowered by the desire for change and ambition for advancement among mortals. Tzeentch\'s true power is sorcery, and as all sorcery flows from the font of the Immaterium, so too is Tzeentch the master of that twisted, chaotic medium of psychic energy.',
+     mainLegion: 'Main legion: Thousand sons'
+    },
+    {name: 'Nurgle',
+     imgUrl: 'img/Nurgle.jpg',
+     altName: 'Plague Lord',
+     warCry: '"Buboes, phlegm, blood and guts! Boils, bogeys, rot and pus! Blisters, fevers, weeping sores! From your wounds the fester pours."',
+     descr: '&emsp;Nurgle, also known as the "Plague Lord," "Grandfather Nurgle," or the "Plague God," among many other honourifics, is the Chaos God of disease, decay, despair, destruction, death and rebirth. In particular, the emotion of despair in mortals empowers the Plague God more than any other. <br><br>&emsp;Nurgle is the Chaos God most directly involved with the plight of mortals, particularly Humans who suffer so acutely from a fear of death, perhaps the oldest fear of that species, or any other. While Nurgle is the god of death and decay, it is also the god of rebirth. Decay is simply one part of the cycle of life, without which no new life could grow.',
+     mainLegion: 'Main legion: Death Guard'
+    },
+    {name: 'Khorne',
+     imgUrl: 'img/Khorne.jpg',
+     altName: 'Lord of Skulls',
+     warCry: '"Blood for the Blood God! Skulls for the Skull Throne!"',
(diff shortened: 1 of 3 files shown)


Tedzury:
Hello. Fixed minor issues during self-check. Feel ready to ask for review :)

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,164 @@
...
+  container.append(firstImg);
+  container.append(text);
+  contentBox.innerHTML = "";
+  contentBox.append(container);
+}
+
+makeFirstPage();
+
+function makeMenu(content) {
+  for (let i = 0; i < content.length; i++) {
```
artem-trubin, line 75: Use `forEach` instead of a regular cycle when possible.

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,164 @@
...
+
+  menuWrapper.append(menuList);
+}
+
+makeMenu(content);
+
+burger.addEventListener("click", menuHandler);
+const menuClasses = menuList.classList;
+
+function menuHandler(e) {
```
artem-trubin, line 96: You can extract `target` right away with `function menuHandler({ target }) { ...`

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,164 @@
...
+    contentBox.addEventListener("click", closeMenuByTap);
+  } else {
+    closeMenu();
+    contentBox.removeEventListener("click", closeMenuByTap);
+  }
+}
+
+const menuItems = document.querySelectorAll(".menu__item");
+
+for (let i = 0; i < content.length; i++) {
```
artem-trubin, line 122: You should use as few event listeners in your work as possible, so please try to use only one for the whole menu.

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,164 @@
...
+  menuItems[i].addEventListener("click", makeContent);
+}
+
+function makeContent(e) {
+  menuHandler();
```
artem-trubin, line 127: I am confused, are you redoing your function for the menu every time you go to another page? Isn't `menuHandler` supposed to receive the `e` parameter (tip: it does but it shouldn't)? Why not run this once somewhere?

artem-trubin (changes requested):
Hi, nice work! Let's make it better.

There is a little problem with functions and their responsibilities. You see, for code to be much more understandable and clearer we should write it as close to its actual meaning as possible. What that means is that if your function is named` makeContent` it is not very clear what exactly is happening there. If we look at your implementation of this function we can see that it does several things:
- it receives the `event` object
- it calls `menuHandler` to reinitiate the menu (which is strange since the menu is static)
- creates a page layout based on the id that is in `event target`
- renders this layout to the page

It's pretty hard to call these actions `Make Content`. 

If we split functions like this into smaller parts that just do their job it becomes much simpler and more clear what is happening in the code. It would be nice to have a function called something like `prepareLayoutById` that gets an ID of a clicked button and **returns** layout. It doesn't need the whole `event` object and it should render to the page right away, it is not its responsibility. You can abstract rendering a prepared layout to a separate function and name it accordingly.

Tedzury:
Hello. Hope I fixed everything needed. Removed for loop, removed unneeded listeners for menu items, renamed function makeContent to prepareLayoutById, split this func to two parts, one to get the Id, second to form actual layout.

Few word about func menuHandler: it is responsible for opening and hiding the menu on mobile devices by burger. If user taps the burger - it opens the menu, same time it attaches the event listener to everything that is not a menu, to make it possible for user to close menu if he/she doesn't want to choose new page, just wanna close menu. In case user taps new page - it closes the menu i renders new content. Kind of switcher to menu, I guess.

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,175 @@
...
+    contentBox.removeEventListener("click", closeMenuByTap);
+  }
+}
+
+menuList.addEventListener("click", prepareLayoutById);
+
+function prepareLayoutById(e) {
+  menuHandler();
+
+  const contentId = getContentId();
```
artem-trubin, line 123: The whole thing with `getContentId` looks strange. You can just do

```js
const contentId = content.findIndex(item => item.name === target.textContent)
```

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,175 @@
...
+
+  function getContentId() {
+    let result;
+    content.forEach((item) => {
+      if (item.name === e.target.textContent) result = content.indexOf(item);
+    });
+    return result;
+  }
+
+  function prepareLayout(contentId) {
```
artem-trubin, line 135: Why does `prepareLayout` has to be a function? You have many of these in code, a function that is just used once, is there a real need for such code to be a function? If you want to group something, maybe just make a comment header.

submissions/Oleksii_Drohachov/DOM_task/script.js
```diff
@@ -0,0 +1,175 @@
...
+    if (contentId != undefined) {
```
artem-trubin, line 136: Look, you have an indentation for the whole block of code just because of this one `if`. You can redo it as a single-line `if` like this:

```js
if (!contentId) return
```

And just continue your code on the same level of indentation.

artem-trubin (changes requested):
Hey, let's make some more improvements!

Tedzury:
Hello! Fixed everything you mentioned, deleted <ul> in html and created it by JS, to only reach the DOM once. 

In last comment  can't use " if (!contentId) return ", because I have an array, and content with index 0 won't be rendered, instead used " if (contentId === undefined) return  "

Made slight refactoring, hope it makes code better for review. 

Also not sure if got you right in last but one comment: " Why does prepareLayout has to be a function? You have many of these in code, a function that is just used once, is there a real need for such code to be a function? If you want to group something, maybe just make a comment header. "

artem-trubin (changes requested):
Hey, that's better! But now you've introduced a new bug: your app breaks if you click inside the menu block, but not on the button. I believe you should add a check in your listener function to check if user actually clicked a button.

Tedzury:
Good evening. Fixed bug, just replaced "if (contentId === undefined) return"  with  "(contentId === -1 ) return;", it happens because of findIndex, it returns -1 if nothing was found, not undefined.

artem-trubin (changes requested):
Good find and a fix for a bug! I still have a question about the `prepareLayout` function though, you use it only once right where you define it, why make it a function at all? Functions are supposed to be some group of code lines and actions to either reuse or use later, and here you can just remove the `function` wrapping and nothing would change.

Tedzury:
Hello. Sorry for my dumbness. It was really unneeded function. Made it just for logical separation. So, now removed it.

artem-trubin (approved):
Nice job!

PR #256, opened 2022-08-22
oop exercise - a-tiny-oop-world

expero1:
### OOP Exercise

[Demo](https://expero1.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/expero1/a-tiny-JS-world/tree/using-oop)

Please, review.

### DIFF
--- a/submissions/expero1/oop-exercise/index.js
+++ b/submissions/expero1/oop-exercise/index.js
@@ -1,0 +1,105 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Habitat {
+  static SPECIES = { DOG: "dog", CAT: "cat", HUMAN: "human" };
+  static GENDER = { MALE: "male", FEMALE: "female" };
+  friends = [];
+  PROPS = ["species", "name", "gender", "legs", "saying"];
+  constructor(species, name, gender, legs, saying) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.saying = saying;
+  }
+
+  addFriend(habitat) {
+    if (!this.friends.includes(habitat)) {
+      this.friends.push(habitat);
+    }
+    if (!habitat.friends.includes(this)) {
+      habitat.friends.push(this);
+    }
+  }
+
+  toString() {
+    let propsKeys = this.PROPS.map((key) => {
+      return this[key];
+    });
+    let info = propsKeys.join("; ");
+    if (this.friends.length > 0) {
+      info += "; ";
+      let friendList = this.friends.reduce((list, friend) => {
+        list.push(friend.species + " " + friend.name);
+        return list;
+      }, []);
+
+      info += friendList.join(", ");
+    }
+    return info;
+  }
+}
+
+class Human extends Habitat {
+  constructor(name, gender, legs, hands, saying) {
+    super(Habitat.SPECIES.HUMAN, name, gender, legs, saying);
+    this.hands = hands;
+    this.PROPS.splice(this.PROPS.indexOf('legs'), 0, 'hands')
+    console.log(this.PROPS);
+  }
+}
+
+class Dog extends Habitat {
+  constructor(name, gender, legs, saying) {
+    super(Habitat.SPECIES.DOG, name, gender, legs, saying);
+  }
+}
+
+class Cat extends Habitat {
+  constructor(name, gender, legs, saying) {
+    super(Habitat.SPECIES.CAT, name, gender, legs, saying);
+  }
+}
+
+
+const dog = new Dog("Sharik", Habitat.GENDER.MALE, 4, "gav-gav!");
+const cat = new Cat("Murka", Habitat.GENDER.FEMALE, 4, "Myau");
+const man = new Human("Igor", Habitat.GENDER.MALE, 2, 2, "Hello, everebody!");
+const woman = new Human("Nata", Habitat.GENDER.FEMALE, 2, 2, "Hi. My name is Nata");
+
+
+dog.addFriend(cat);
+dog.addFriend(cat);
+man.addFriend(cat);
+man.addFriend(woman);
+
+// ======== OUTPUT ========
+[dog, cat, man, woman].forEach((habitat) => {
+  print(habitat);
+});
+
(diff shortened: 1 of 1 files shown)


submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    if (!habitat.friends.includes(this)) {
+      habitat.friends.push(this);
+    }
+  }
+
+  toString() {
+    let propsKeys = this.PROPS.map((key) => {
+      return this[key];
+    });
```
OleksiyRudenko, line 36: We do not need curly braces and `return` keyword when arrow function body has the return statement only.
```
.map((v) => {
  return a[v];
})

.map((v) => a[v])

.map(v => a[v]) // parentheses around arrow function params can also be omitted when parameters count is exactly and no destructuring involved. Also depends on teams' code style conventions

```

submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+    let info = propsKeys.join("; ");
+    if (this.friends.length > 0) {
+      info += "; ";
+      let friendList = this.friends.reduce((list, friend) => {
+        list.push(friend.species + " " + friend.name);
+        return list;
+      }, []);
+
+      info += friendList.join(", ");
+    }
```
OleksiyRudenko, line 46: A use case for `Array#map`, as here we convert every single array element into correlating data. You won't need extra variable. `join` here would be perfectly chainable.

Also, if this conversion is done before `let info = propsKeys.join("; ");` and its result is made a part of `propKeys` array then `propsKeys.join("; ")` will join all array elements, including the list of friends.

Perhaps, you will be able to get rid of `info` variable as well.

submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  addFriend(habitat) {
+    if (!this.friends.includes(habitat)) {
+      this.friends.push(habitat);
+    }
+    if (!habitat.friends.includes(this)) {
+      habitat.friends.push(this);
+    }
+  }
```
OleksiyRudenko, line 31: Nice solution to make friends mutual.

submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Habitat {
+  static SPECIES = { DOG: "dog", CAT: "cat", HUMAN: "human" };
+  static GENDER = { MALE: "male", FEMALE: "female" };
```
OleksiyRudenko, line 13: Good decision to make domain-specific constants a static part of the root class.

submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return info;
+  }
+}
+
+class Human extends Habitat {
+  constructor(name, gender, legs, hands, saying) {
+    super(Habitat.SPECIES.HUMAN, name, gender, legs, saying);
+    this.hands = hands;
+    this.PROPS.splice(this.PROPS.indexOf('legs'), 0, 'hands')
```
OleksiyRudenko, line 55: Child classes shouldn't rely on base class internals. Imagine, the hierarchy is big, split across files and someone is tasked to rename `legs` on the base class into something else. It will take time to find a bug.
Root class could provide mechanics to group any types of limbs, since we want them to stay close.
This would probably lead to treat compound object properties as a special case, just like friends that are a special case of array property.

OleksiyRudenko (changes requested):
@expero1 very well done!
We still can leverage specialized `Array` methods.
Also handling not own properties by classes (child's prop by base class and base class prop by child class or indirectly referring to them via harcoded values) is an anti-pattern. Solvable within OOP/SOLID paradigm.
Let me know if you want to take this challenge.

submissions/expero1/oop-exercise/index.js
expero1: Thank you for your comment. You are right, after the refactoring, the code is much cleaner and easier to read.

submissions/expero1/oop-exercise/index.js
expero1: Your truth. I must use the simplified syntax as it is easier to read. I try to accustom myself to it, but I do not always notice cases when it can be used.

submissions/expero1/oop-exercise/index.js
expero1: Thanks for the important note. I tried to do as you recommended. I hope it worked out exactly as you said. This is a very important experience for me in writing code. Since it is one thing to know the basics of a programming language. And it’s a completely different thing when you have to apply certain patterns and avoid typical mistakes that are now invisible, but in the future can lead to errors that will take a long time to look for. Thanks again, with your help I've gotten a little better at coding.

submissions/expero1/oop-exercise/index.js
expero1: Thanks.

submissions/expero1/oop-exercise/index.js
expero1: Thanks. I try to use this approach in my code because it allows for more controlled code with fewer bugs.

expero1:
Thank you. Your code review was very helpful and educational for me. I will try to further study the principles of SOLID in order to correctly apply them in practice.

submissions/expero1/oop-exercise/index.js
```diff
@@ -0,0 +1,123 @@
...
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Habitat {
+  static SPECIES = { DOG: "dog", CAT: "cat", HUMAN: "human" };
+  static GENDER = { MALE: "male", FEMALE: "female" };
+  static LIMBS = { HANDS: "hands", LEGS: "legs" };
```
OleksiyRudenko, line 14: Base class doesn't need to provide all sorts of limbs for all possible situations of child classes.
Why we cann't treat any limbs as any other properties like `species` or `name`?
If it is a shared property it is then defined at some shared (base) class. If a prop unique to a specific class so let that specific class define and handle that property.
Adding a class-specific property shouldn't affect any of its parents or children.
This what Open-Closed and encapsulation principles are.

OleksiyRudenko (changes requested):
@expero1 well done.
However we need to make properties handling OOP and SOLID compliant.
What if we want dozens of different types of limbs for different species? How cluttered the base class would become?

OleksiyRudenko:
Up

PR #259, opened 2022-08-22
Tiny js world

Tedzury:
### Tiny_JS_world

[Demo](https://tedzury.github.io/Tiny_JS_world/) |
[Code base](https://github.com/Tedzury/Tiny_JS_world)

Please, review.

### DIFF
--- a/submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
+++ b/submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
@@ -1,82 +1,58 @@
-/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
-   Complete the below for code reviewers' convenience:
+const dog = {
+  species: "dog",
+  name: "Barbos",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+  saying: "gav-gav kaje pes",
+  friends: ["definitely everybody in the world !"],
+};
 
-   Code repository: _put repo URL here_
-   Web app: _put project's github pages URL here_
-   */
+const cat = {
+  species: "cat",
+  name: "Sonya",
+  gender: "female",
+  legs: 4,
+  hands: 0,
+  saying: "meow-meow, skinbag...",
+  friends: ["definitely nobody except her own slaves"],
+};
 
-// ======== OBJECTS DEFINITIONS ========
-// Define your objects here
-catsPhrase = "meow-meow skinbag";
+const man = {
+  species: "human",
+  name: "Oleksii",
+  gender: "male",
+  legs: 2,
+  hands: 2,
+  saying: "Wanna be frontend ninja in future!",
+  friends: ["Alex", "Victoria", "every lovely kottan on the course"],
+};
 
-const tinyWorldInhabitants = [
-  {
-    species: "dog",
-    name: "Barbos",
-    gender: "male",
-    legs: 4,
-    hands: 0,
-    saying: "gav-gav kaje pes",
-    friends: ["definitely everybody in the world !"],
-  },
-  {
-    species: "cat",
-    name: "Sonya",
-    gender: "female",
-    legs: 4,
-    hands: 0,
-    saying: catsPhrase,
-    friends: ["definitely nobody except her own slaves"],
-  },
-  {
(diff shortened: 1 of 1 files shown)


Tedzury:
Hello! Fixed using of object.values, replaced it for explicit list of values.  Humbly asking for review :)

submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+      for (let i = 0; i < outputArr.length; i++) {
+         if (Array.isArray(outputArr[i])) {
+            outputArr[i] = outputArr[i].join(', ')
+         }
+      }
+
+      const result = outputArr.join('; ');
+      
+      return result
```
OleksiyRudenko, line 65: (1) As a lazy peer developer tasked to add 20 more properties I'd like to those props names without need to write `obj.` every time.

(2) We can `.map` through array of initial data, whatever it is, converting arrays to strings

(3) As after (2) we have only strings and numbers in our array we can chain conversion `.map` with `.join`

(4) ...and finally we can have a single expression and do not need `result` variable.

submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
```diff
@@ -0,0 +1,72 @@
...
+   }
+   
+   tinyWorldInhabitants.forEach( item => print(formOutput(item)));
+
+
+
+
```
OleksiyRudenko, line 72: Tune your IDE to keep one and only one empty line at the end of file.
Also use prettier.io plugin to do code formatting job for you.

OleksiyRudenko (changes requested):
@Tedzury good job! Let's polish it.
(1) use JS semicolons consistently across your codebase;
(2) see the comments below

Tedzury:
Good morning! Tried to fix my code, hope it's fine. 

To be honest, after an hour of tries to get it to work, I cheeted and looked up for finished works of my teemmates in kottans repo. The problem was in the print( ) function. I totally forgot that forEach( ) does not return anything, and desperately tried to return an array for me, to work further, but if I call print inside the forEach, I don't even need the array to form the output, just used join('; ').

Anyway, I'm sorry for that, shame on me :(

submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
```diff
@@ -0,0 +1,101 @@
...
+  "saying",
+  "friends",
+];
+
+tinyWorldInhabitants.forEach((person) => {
+  print(
+    keys
+      .map((key) => {
+        if (Array.isArray(person[key])) {
+          person[key] = person[key].join(", ");
```
OleksiyRudenko, line 95: This way you change the person's property on spot. What if another fragment of code would like to treat the list differently? We only need this transformation be scoped exclusively for this fragment of code purpose.

You may leverage a ternary operator here. This will also eliminate need of `return` keyword and function body wrapping with `{}`

submissions/Oleksii_Drohachov/Tiny_JS_world/index.js
```diff
@@ -0,0 +1,101 @@
...
+    saying: "When will you bring the money to home, honey?",
+    friends: ["Natasha", "Katya", "Lada"],
+  },
+  {
+    species: "catWoman",
+    name: "Anjela",
+    gender: "female",
+    legs: 2,
+    hands: 2,
+    saying: catsPhrase,
```
OleksiyRudenko, line 55: This approach won't ensure catWoman's saying when cat's will change.
You do not need to fix this here. But you will need different approach when you are back to this JS World after the next task.

OleksiyRudenko (changes requested):
@Tedzury well done!
Let's polish it a bit.

Tedzury:
Good evening! Fixed all issues, changed 'if' for ternary operator, and set a cat as a prototype for catWoman, and for this reason slightly refactored array of characters, to set a proto.

Also want to thank you very much for everything you do for students with that course. I really appreciate that.

OleksiyRudenko (approved):
@Tedzury job well done!

PR #286, opened 2022-08-24
add frogger game

YuliiaDikun:
### Classic Frogger Game

[Demo](https://yuliiadikun.github.io/frontend-nanodegree-arcade-game/) |
[Code base](https://github.com/YuliiaDikun/frontend-nanodegree-arcade-game/blob/master/js/app.js)

Please, review.

### DIFF
--- a/submissions/dikun/frogger-game/app.js
+++ b/submissions/dikun/frogger-game/app.js
@@ -13,8 +13,8 @@ const GAME_FIELD = {
 };
 
 const PLAYER_START = {
-	x: 202,
-	y: 400
+	x: CELL.width * 2.02,
+	y: CELL.height * 4.7
 };
 
 const ENEMY_SPEED = {
@@ -27,16 +27,22 @@ const SPRITE = {
 	player: 'images/char-boy.png'
 };
 
-const SPEED = () => ENEMY_SPEED.min + Math.floor(Math.random() * ENEMY_SPEED.max);
+const speed = () => ENEMY_SPEED.min + Math.floor(Math.random() * ENEMY_SPEED.max);
 const OVERLAP_AMOUNT = 70;
 const BORDER = 20;
 
-const Enemy = function (x, y, SPEED, player) {
+const LINE = {
+	1: CELL.height - BORDER,
+	2: CELL.height * 2 - BORDER,
+	3: CELL.height * 3- BORDER
+};
+
+const Enemy = function (x, y, speed, player) {
 	this.x = x;
 	this.y = y;
-	this.SPEED = SPEED();
+	this.speed = speed();
 	this.player = player;
-	this.SPRITE = SPRITE.enemy;	
+	this.sprite = SPRITE.enemy;	
 };
 
 Enemy.prototype.checkColission = function () {
@@ -51,21 +57,21 @@ Enemy.prototype.checkColission = function () {
 
 Enemy.prototype.update = function (dt) {
 	if (this.x > GAME_FIELD.right) {
-		this.x -= GAME_FIELD.right;
-		this.SPEED = SPEED();
+		this.x -= GAME_FIELD.right + CELL.width;
+		this.speed = speed();
 		}
-	this.x += this.SPEED * dt;
+	this.x += this.speed * dt;
 	this.checkColission();
 };
 
 Enemy.prototype.render = function () {
-	ctx.drawImage(Resources.get(this.SPRITE), this.x, this.y);
+	ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 };
 
 const Player = function () {
 	this.x = PLAYER_START.x;
 	this.y = PLAYER_START.y;
-	this.SPRITE = SPRITE.player;	
+	this.sprite = SPRITE.player;	
 };
 
 Player.prototype.update = function (dt) {};
@@ -91,13 +97,22 @@ Player.prototype.handleInput = function (keyPress) {
 };
 
 Player.prototype.render = function () {
-	ctx.drawImage(Resources.get(this.SPRITE), this.x, this.y);
+	ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 };
 
 const player = new Player(PLAYER_START.x, PLAYER_START.y);
 
-const allEnemies = [CELL.height - BORDER, CELL.height*2 - BORDER, CELL.height*3 - BORDER]
-				   .map(value => {return new Enemy(GAME_FIELD.left, value, SPEED, player);});
+const allEnemies = [];
+
+function createEnemyPerLine (line, numberOfEnemies) {	
+	for (let i = 0; i < numberOfEnemies; i++) {		
+		allEnemies.push(new Enemy((GAME_FIELD.left - CELL.width), LINE[line], speed, player));		
+	}
+}
+
+createEnemyPerLine(1, 1);
+createEnemyPerLine(2, 2);
+createEnemyPerLine(3, 1);
 
 // This listens for key presses and sends the keys to your
 // Player.handleInput() method. You don't need to modify this.
--- a/submissions/dikun/frogger-game/app.js
+++ b/submissions/dikun/frogger-game/app.js
@@ -1,20 +1,20 @@
 "use srtict";
 
 const CELL = {
-	width: 100,
-	height: 85
+	width: 101,
+	height: 80
 };
 
 const GAME_FIELD = {
 	top: 0,
 	right: CELL.width * 5,
-	bottom: CELL.height * 5,
+	bottom: CELL.height * 6,
 	left: 0
 };
 
 const PLAYER_START = {
-	x: CELL.width * 2.02,
-	y: CELL.height * 4.7
+	x: CELL.width * 2,
+	y: CELL.height * 5 
 };
 
 const ENEMY_SPEED = {
@@ -30,12 +30,6 @@ const SPRITE = {
 const speed = () => ENEMY_SPEED.min + Math.floor(Math.random() * ENEMY_SPEED.max);
 const OVERLAP_AMOUNT = 70;
 const BORDER = 20;
-
-const LINE = {
-	1: CELL.height - BORDER,
-	2: CELL.height * 2 - BORDER,
-	3: CELL.height * 3- BORDER
-};
 
 const Enemy = function (x, y, speed, player) {
 	this.x = x;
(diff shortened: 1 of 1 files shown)


YuliiaDikun:
I have completed self-checks and fixed the code accordingly. Please, look at it. Thanks for your time.

OleksiyRudenko:
Also please before submitting your next task make your code editor do indents with 2 spaces, not tabs.
Tabs expand in different ways on different environments. 8 spaces indentation at times makes reading and following code structure a bit painful.

Do not do this for this task as it will mark the entire code as changed.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
+"use srtict";
+
+const cell = {
+	width: 100,
+	height: 85
+};
```
OleksiyRudenko, line 6: By convention, constants are named all caps and snake-case (e.g. `CELL`, `GAME_FIELD`). This makes them outstand in the code and let the reader instantly know they are constants and not just global variables.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
...
+	max: 250 
+}; 
+
+const sprite = {
+	enemy: 'images/enemy-bug.png',
+	player: 'images/char-boy.png'
+};
+
+const speed = () => enemySpeed.min + Math.floor(Math.random() * enemySpeed.max);
+const crash = 70;
```
OleksiyRudenko, line 31: As a peer developer I want to understand what does `crash` mean and how do I use it. Is this power of crash, overlap by one of axis, overlap amount or anything else?

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
...
+
+	this.checkColission = function () {
+		if(this.player.x + crash > this.x &&
+			this.player.x < this.x + crash &&
+			this.player.y + crash > this.y &&
+			this.player.y < this.y + crash) {
+			this.player.x = playerStart.x;
+			this.player.y = playerStart.y;
+		}
+	};
```
OleksiyRudenko, line 49: This approach defines a **copy** of function per every instance of class.
We do not want this. One of the crucial points of prototypical inheritance model is to have a single copy of method per class, stored in its prototype object.
Check the original project code for a pattern to follow.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
...
+			alert('You win!');	
+		}			
+	};
+};
+
+Function.prototype.construct = function (position) {
+	let gameEnteties = Object.create(this.prototype);
+	this.apply(gameEnteties, position);
+	return gameEnteties;
+};
```
OleksiyRudenko, line 102: We do not want to play around with JS built in classes/objects.
If you want a base class, you need a class function like `Player` or `Enemy` and then you bind `Player` and `Enemy` classes with this base class by `Object.create`.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
...
+
+const playerPosition = [playerStart.x, playerStart.y];
+const player = Player.construct(playerPosition);
```
OleksiyRudenko, line 105: Adjust the approach.
An instance would be created with the `new` keyword.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,123 @@
...
+
+const allEnemies = [cell.height - border, cell.height*2 - border, cell.height*3 - border].map((value) => {
+	const enemyPosition = [gameField.left, value, speed, player];
+	return Enemy.construct(enemyPosition);
```
OleksiyRudenko, line 109: The composition of an array to the right doesn't look like a position altogether.
Also class functions are quite normal functions and can have normal named parameters.

OleksiyRudenko (changes requested):
@YuliiaDikun a pretty decent job!
But something went sideways with class methods declaration and building a base class.

Feel free adding changes incrementally and asking questions.

I also wonder where did you pick this class methods definition pattern?

submissions/dikun/frogger-game/app.js
YuliiaDikun: but I think I can't use the new keyword...(following "Requirements re OOP:" last point)

OleksiyRudenko: Classes defined with `prototype`-based syntax are instantiated with `new` keyword. It is way older than the `class` syntax.

The referred requirement is about class **extension**.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,113 @@
...
+const SPRITE = {
+	enemy: 'images/enemy-bug.png',
+	player: 'images/char-boy.png'
+};
+
+const SPEED = () => ENEMY_SPEED.min + Math.floor(Math.random() * ENEMY_SPEED.max);
+const OVERLAP_AMOUNT = 70;
+const BORDER = 20;
+
+const Enemy = function (x, y, SPEED, player) {
```
OleksiyRudenko, line 34: Function parameters (e.g. `SPEED`) are not constants anymore. We may send constants to functions, but the function itself doesn't really care where the parameter value comes from.

YuliiaDikun:
I return to the original inheritance pattern and use a `new` keyword. Also, change my const to uppercase and crash to OVERLAP_AMOUNT. Please, look at the changes! Thank you for all your remarks and comments.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,113 @@
...
+};
+
+Player.prototype.render = function () {
+	ctx.drawImage(Resources.get(this.SPRITE), this.x, this.y);
+};
+
+const player = new Player(PLAYER_START.x, PLAYER_START.y);
+
+const allEnemies = [CELL.height - BORDER, CELL.height*2 - BORDER, CELL.height*3 - BORDER]
+				   .map(value => {return new Enemy(GAME_FIELD.left, value, SPEED, player);});
```
OleksiyRudenko, line 100: (1) As a peer developer I am assigned a task to make bugs elegantly flow in from outside. As gracefully as they leave it. Intuition says I need only change the first argument of Enemy constructor (`GAME_FIELD.left` gives a good hint on it). If I go and change the constant value, probably something else will break.
What constant can I subtract from the field left border coord to achieve this? Can you please do this for me?

(2) Arrow functions have a nice feature. If its body has a single statement and it is `return` then function body braces and `return` keyword can be omitted.
```
.map(param => { return param * 2; })

.map(param => param * 2)
```
This adds to code readability.

(3) `value` is to generic word. What this parameter represents in the game domain/world?

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,113 @@
...
+const GAME_FIELD = {
+	top: 0,
+	right: CELL.width * 5,
+	bottom: CELL.height * 5,
+	left: 0
+};
+
+const PLAYER_START = {
+	x: 202,
+	y: 400
```
OleksiyRudenko, line 17: I feel these can be calculated.

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,113 @@
...
+	}		
+};
+
+Player.prototype.render = function () {
+	ctx.drawImage(Resources.get(this.SPRITE), this.x, this.y);
+};
+
+const player = new Player(PLAYER_START.x, PLAYER_START.y);
+
+const allEnemies = [CELL.height - BORDER, CELL.height*2 - BORDER, CELL.height*3 - BORDER]
```
OleksiyRudenko, line 99: How can we reduce repetition here? Ideally I'd like to add a row number and call it a day, if I need to add yet another bug.

OleksiyRudenko (changes requested):
@YuliiaDikun great improvement to code comprehension!
Let's polish it.

YuliiaDikun:
Please, see my changes. Thank you for your time!

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,128 @@
...
+
+function createEnemyPerLine (line, numberOfEnemies) {	
+	for (let i = 0; i < numberOfEnemies; i++) {		
+		allEnemies.push(new Enemy((GAME_FIELD.left - CELL.width), LINE[line], speed, player));		
+	}
+}
+
+createEnemyPerLine(1, 1);
+createEnemyPerLine(2, 2);
+createEnemyPerLine(3, 1);
```
OleksiyRudenko, line 115: Mapping was much cleaner and readable.
Now we have longer code and also repetition.
What if I want to add 20 more rows? How can I avoid adding 20 more lines of code?
Take a step back and think how could you optimize `[CELL.height - BORDER, CELL.height*2 - BORDER, CELL.height*3 - BORDER]`. There is a pattern there. You must see it better if the line of code modified as follows: `[CELL.height*1 - BORDER, CELL.height*2 - BORDER, CELL.height*3 - BORDER]`.

Think of chained mapping also look into what parameters a callback for `map` receives and you could leverage.

Examples:

```
[1, 5, 7, 3].map(heightMeters => heightMeters * 100).map((heightCentimeters, numberOfWalls) => buildHouseFragment(heightCentimeters, numberOfWalls + 1));
```

Of course the above is optimizable. We can just multiply source data by 100 when calling `buildHouseFragment`. This is specifically to make you start thinking of what numbers like 1, 2, 3 in your original fragment mean.

OleksiyRudenko (changes requested):
@YuliiaDikun nice try.
We need to make a step back and look into another direction.

submissions/dikun/frogger-game/app.js
OleksiyRudenko: Now I can **read** that initial arrays contains lines. That exaplains a lot to me as a peer developer and now I clearly understand what to do if I want to add yet another enemy to run across line 2 for example.
Although I'd suggest rename `line` as `rowNumber` or `laneNumber` or `trackNumber`

submissions/dikun/frogger-game/app.js
```diff
@@ -0,0 +1,128 @@
...
+const GAME_FIELD = {
+	top: 0,
+	right: CELL.width * 5,
+	bottom: CELL.height * 5,
+	left: 0
+};
+
+const PLAYER_START = {
+	x: CELL.width * 2.02,
+	y: CELL.height * 4.7
```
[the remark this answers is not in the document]
YuliiaDikun: Fixed.

submissions/dikun/frogger-game/app.js
YuliiaDikun: I spent a few hours trying to understand how I can add a new bug in line 2. OMG. Thank you a lot for all your remarks. Hope it is fixed now too.

YuliiaDikun:
Please. see my changes. Thanks!

OleksiyRudenko (approved):
@YuliiaDikun excellent job!

PR #162, opened 2022-08-15
HTML & CSS Practice

martthaa:
### HTML & CSS Practice

[Demo](https://martthaa.github.io/sweetPopUp/) |
[Code base](https://github.com/martthaa/sweetPopUp)

The code is submitted in a dedicated feature branch.

Only code files are submitted.

PR #300, opened 2022-08-25
HTML&CSS Practise PopUp

martthaa:
### HTML&CSS Practise

[Demo](https://martthaa.github.io/sweetPopUp/) |
[Code base](https://github.com/martthaa/sweetPopUp)

### DIFF
--- a/submissions/martthaa/PopUpTask/index.html
+++ b/submissions/martthaa/PopUpTask/index.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="author" content="Marta Sobotyk"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Sweet PopUp</title>
+    <link rel="shortcut icon" href="./src/img/dessert_donut_icon.png">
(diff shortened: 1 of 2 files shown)


OleksiyRudenko:
Is it a duplicate of #162 ?

martthaa:
Yes

OleksiyRudenko:
Why do we need this PR?

martthaa:
I had problems with adding files, I have pushed to github folder, but then I  see a white arrow on my folders and when i click on the folder, it does not open it. On my local machine, it has contents, but in github i cannot access them. Well, I tried different commands that I found on the Internet about this, but the problem was not fixed and I decided to add files through the interface, but the mentor asked me to solve the problem with the file, I did not succeed, so I decided to add a new branch, where I created a new PR.

 Now I understand why I got this file, my mistake was that I just past the folder (which I previously pushed into GitHub to create Demo), that is not correct, because in fact I wanted to create a repository inside a repository. Thus was created gitlink - nested Git repository

OleksiyRudenko:
Can we close this PR or do we still need it?

martthaa:
I think to close

martthaa:
sorry, that I distracted you with this PR

OleksiyRudenko:
Closing this as the task is already in review under #162

OleksiyRudenko:
@martthaa no worries. We figured this out

PR #303, opened 2022-08-25
html-css-popup

AaLin-Git:
### Hooli-style Popup

[Demo](https://aalin-git.github.io/Hooli-style-Popup/) |
[Code base](https://github.com/AaLin-Git/Hooli-style-Popup)

Please, review.

### DIFF
--- a/submissions/AaLin-Git/html-css-popup/index.html
+++ b/submissions/AaLin-Git/html-css-popup/index.html
@@ -24,7 +24,7 @@
                 <img class="menuIcon" src="./img/menu.svg" alt="menu" />
               </label>
 
-              <section class="menuPopup">
+              <div class="menuPopup">
                 <ul class="menuPopupList">
                   <li class="popupLink">
                     <a class="menuPopupItem" href="#">
@@ -102,8 +102,11 @@
                   </li>
                 </ul>
                 <input type="checkbox" id="menuInputSub" class="input" />
-                <label class="labelSub more" for="menuInputSub">More </label>
-                <section class="menuPopupSub">
+                <label class="labelSub" for="menuInputSub">
+                  <p class="moreP">More</p>
+                </label>
+                <p class="lessP">Less</p>
+                <div class="menuPopupSub">
                   <ul class="menuPopupList">
                     <li class="popupLink">
                       <a class="menuPopupItem" href="#">
@@ -196,8 +199,8 @@
                   <a class="more" href="https://github.com/kottans"
                     >Even more from Kottans</a
                   >
-                </section>
-              </section>
+                </div>
+              </div>
             </li>
             <li>
               <a class="labelIcon" href="#">
--- a/submissions/AaLin-Git/html-css-popup/index.html
+++ b/submissions/AaLin-Git/html-css-popup/index.html
@@ -19,10 +19,11 @@
               <a class="contentItem" href="#">Images</a>
             </li>
             <li class="hiddenNavbar contentLink">
+              <input type="checkbox" id="menuInput" class="input" />
               <label for="menuInput" class="labelIcon">
                 <img class="menuIcon" src="./img/menu.svg" alt="menu" />
               </label>
-              <input type="checkbox" id="menuInput" class="input" />
+
               <section class="menuPopup">
                 <ul class="menuPopupList">
                   <li class="popupLink">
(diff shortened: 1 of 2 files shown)


AaLin-Git:
Hi! I added a focus to the labels, replaced <p> with <span>, added a cursor and optimized a css code. I think now everything is correct, well I hope so :)

submissions/AaLin-Git/html-css-popup/style.css
```diff
@@ -0,0 +1,270 @@
...
+
+.content {
+  margin: 15px 30px;
+  flex-direction: row;
+  justify-content: space-between;
+  width: 50%;
+  list-style: none;
+}
+
+.hiddenNavbar section,
```
A-Ostrovnyy, line 162: Avoid styling by HTML tags except if you add base styles. [Reason](https://frontstuff.io/you-need-to-stop-targeting-tags-in-css) Also, avoid too specific CSS selectors. [Read about that](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity)

submissions/AaLin-Git/html-css-popup/index.html
```diff
@@ -0,0 +1,223 @@
...
+            <li>
+              <a class="contentItem" href="#">Images</a>
+            </li>
+            <li class="hiddenNavbar">
+              <input type="checkbox" id="menuInput" class="input" />
+              <label for="menuInput" class="labelIcon">
+                <img class="menuIcon" src="./img/menu.svg" alt="menu" />
+              </label>
+
+              <section class="menuPopup">
```
A-Ostrovnyy, line 27: Check your markup in [validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Faalin-git.github.io%2FHooli-style-Popup%2F)

submissions/AaLin-Git/html-css-popup/index.html
```diff
@@ -0,0 +1,223 @@
...
+                        class="menuPopupIcon"
+                        src="./img/neptune.svg"
+                        alt=""
+                      />
+                      <span class="textItem">Neptune</span>
+                    </a>
+                  </li>
+                </ul>
+                <input type="checkbox" id="menuInputSub" class="input" />
+                <label class="labelSub more" for="menuInputSub">More </label>
```
A-Ostrovnyy, line 105: Logically this button must be hidden when the second part of the popup is open.

A-Ostrovnyy (changes requested):
Nice design!

submissions/AaLin-Git/html-css-popup/style.css
AaLin-Git: Thanks, I removed section. Should I  also remove * and htms tags from the reset CSS?

submissions/AaLin-Git/html-css-popup/index.html
AaLin-Git: Replaced the sections with the divs

submissions/AaLin-Git/html-css-popup/index.html
AaLin-Git: I changed button more to less, when it is pressed :)

AaLin-Git:
I think now it's correct :)

submissions/AaLin-Git/html-css-popup/style.css
A-Ostrovnyy: No, in reset.css you can use HTML tags because you set basic styles in that file.

submissions/AaLin-Git/html-css-popup/index.html
A-Ostrovnyy: From a UX perspective, it's not ok because users see a list of links then they see the button and after this button, they see the second part of the popup. What if you have a couple of parts of links inside the popup, you will add a button before each of them or only before the last one?

AaLin-Git: It depends on the purpose. If this is a header menu, then it is better to add one button before the last part to be able to collapse the extended menu to the main one. And if this is a menu filter for an online store, for example, then I would make a popup for each section.

AaLin-Git:
I removed button more, when it's pressed

A-Ostrovnyy (approved):
Well done!

PR #339, opened 2022-08-29
OOP exercise

crealime:
### OOP exercise

[Demo](https://crealime.github.io/a-tiny-JS-world) |
[Code base](https://github.com/crealime/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/crealime/oop-exercise/index.js
+++ b/submissions/crealime/oop-exercise/index.js
@@ -7,6 +7,16 @@
 
 // ======== OBJECTS DEFINITIONS ========
 
+const pronounsHisHer = {
+  male: 'His',
+  female: 'Her'
+}
+
+const pronounsHeShe = {
+  male: 'He',
+  female: 'She'
+}
+
 function getRandomId() {
   return  new Date().getTime() + '_' + Math.floor(10000 + Math.random() * (99999 + 1 - 10000))
 }
@@ -27,15 +37,15 @@ class Inhabitant {
   }
 
   getAcquaintanceInfo() {
-    return `This is ${this.name} and ${this.gender === 'male' ? 'he' : 'she'} is a ${this.species}.`
+    return `This is ${this.name} and ${pronounsHeShe[this.gender].toLowerCase()} is a ${this.species}.`
   }
 
   getGenderInfo() {
-    return `${this.gender === 'male' ? 'His' : 'Her'} gender is ${this.gender}.`
+    return `${pronounsHisHer[this.gender]} gender is ${this.gender}.`
   }
 
   getLimbsInfo() {
-    return `${this.gender === 'male' ? 'He' : 'She'} have ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
+    return `${pronounsHeShe[this.gender]} have ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
   }
 
   getFriendsInfo() {
@@ -43,7 +53,7 @@ class Inhabitant {
   }
 
   getSayingInfo() {
-    return `${this.gender === 'male' ? 'He' : 'She'} usually says hello like this: ${this.saying}`
+    return `${pronounsHeShe[this.gender]} usually says hello like this: ${this.saying}`
   }
 
   getFullInfo() {
@@ -79,7 +89,7 @@ class Human extends Inhabitant {
   }
 
   getLimbsInfo() {
-    return `${this.gender === 'male' ? 'He' : 'She'} have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
+    return `${pronounsHeShe[this.gender]} have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
   }
 
   sayFullInfo() {
--- a/submissions/crealime/oop-exercise/index.js
+++ b/submissions/crealime/oop-exercise/index.js
@@ -66,9 +66,18 @@ class CatWoman extends Inhabitant {
   }
 
   getFullName() {
-    return `${this.name} ${this.surname}`
+    return `${super.getFullName()} ${this.surname}`
   }
 }
+
+const fullInfo =  {
+  sayFullInfo() {
+    return `${this.saying} My name is ${this.getFullName()} and I am a ${this.species}. My gender is ${this.gender}. I have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}. I have ${this.friends.length} ${this.friends.length === 1 ? 'friend' : 'friends'}${this.friends.length > 0 ? ': ' + this.friends.map(friend => friend.name).join(', ') : ''}.`
+  }
+}
+
+Object.assign(Human.prototype, fullInfo)
+Object.assign(CatWoman.prototype, fullInfo)
 
 const dog = new Dog('Marley', 'male', 4)
 const cat = new Cat('Sophie', 'female', 4)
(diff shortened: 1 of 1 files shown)


crealime:
I made corrections based on the checklist. 
I have a question. Is this code can be considered an example of composition? 
`super('cat-woman', name, 'female', new Cat().saying)`
We also create an object here and add its property to another object. Can this approach be considered good practice?

OleksiyRudenko:
> I have a question. Is this code can be considered an example of composition? `super('cat-woman', name, 'female', new Cat().saying)` We also create an object here and add its property to another object. Can this approach be considered good practice?

Not really. Exactly because we create an instance to just use a single property of it. Besides, we will use a **copy** of the property, so newly created object will just clutter memory a bit. But imagine we create 1000's of cat-women.

Look into static lass properties and methods. Remember, primitives are copied, methods are object that are passed by reference. And we need to create a stron tie between cat-women and cats in regards to sayings. So, when cat typical saying changes further in the code, all cat-women automagically start saying the same.

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,143 @@
...
+
+function getFullInfo(inhabitant) {
+  const getAcquaintance = `This is ${inhabitant.getFullName()} and ${inhabitant.gender === 'male' ? 'he' : 'she'} is a ${inhabitant.species}.`
+  const getGender = `${inhabitant.gender === 'male' ? 'His' : 'Her'} gender is ${inhabitant.gender}.`
+  const getLimbs = `${inhabitant.gender === 'male' ? 'He' : 'She'} have ${inhabitant.hands || 0} ${inhabitant.hands === 1 ? 'hand' : 'hands'} and ${inhabitant.legs || 0} ${inhabitant.legs === 1 ? 'leg' : 'legs'}.`
+  const getFriends = `${inhabitant.name} have ${inhabitant.friends.length} ${inhabitant.friends.length === 1 ? 'friend' : 'friends'}${inhabitant.friends.length > 0 ? ': ' + inhabitant.friends.map(friend => friend.name).join(', ') : ''}.`
+  const getSaying = `${inhabitant.gender === 'male' ? 'He' : 'She'} usually says hello like this: ${inhabitant.saying}`
+
+  return `${getAcquaintance} ${getGender} ${getLimbs} ${getFriends} ${getSaying}`
+}
```
OleksiyRudenko, line 127: Are there any good reasons not have all this handled by classes? (not a rhetoric question)
Imagine, we add fish, birds, octopi, and spiders. How big this function would become? (a rhetoric question)
Separation of concerns and encapsulation are good principles to follow.

There must be good reasons to handle full name in one way and other props in a completely different way. What these reasons are?

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,143 @@
...
+
+allInhabitant.forEach(inhabitant => {
+  print(getFullInfo(inhabitant))
+})
+
+function getMembersOfSociety() {
+  const names = isInSociety(secretSociety, allInhabitant)
+  let textAfterNames = ''
+  if (names.length === 1) textAfterNames = 'in a secret society...'
+  if (names.length > 1) textAfterNames = 'are in a secret society...'
```
OleksiyRudenko, line 137: What are other possible outcomes?
Would a ternary operator do the job?

(both questions aren't rhetoric)

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,143 @@
...
+function addToSociety(society, member) {
+  society.push(member.id)
+}
+
+addToSociety(secretSociety, catWoman)
+addToSociety(secretSociety, pirate)
+
+function isInSociety(society, members) {
+  return members.filter(member => society.includes(member.id)).map(member => member.name)
+}
```
OleksiyRudenko, line 98: Looks like a secret society deserves to be a class. Just for the sake of scoping and encapsulation of storage and behaviour.
Optional: think of Singleton pattern.

OleksiyRudenko (changes requested):
@crealime a really good job.
Yet there is some room for improvement.

submissions/crealime/oop-exercise/index.js
crealime: Yes, at first I thought that not all residents can tell about themselves. Then I decidet that there can be many such functions in the world, some functions can be created dynamically. Therefore, I did not put it in classes. This feature is quite flexible, but it can be added functionality for those inhabitants, who do not make sounds at all, such as fish. Мaybe gender for hermaphrodites.. However, to practice, I added this mixin to Human and CatWoman classes. 

I made getFullName method because some inhabitants  have a first and last name, while others have only first name. I thought it was more useful.

submissions/crealime/oop-exercise/index.js
crealime: Thanks! I did not understand right away. The array may be empty. Ternary operator, really more convenient. Done.

submissions/crealime/oop-exercise/index.js
crealime: Done.

crealime:
Thanks for review and advice. I've made changes. Details in the comments.

submissions/crealime/oop-exercise/index.js
OleksiyRudenko: Composition would work like this. Mixing OOP and composition is not the best practice.
`getFullName` is valid case though. And this is the key -- when behaviour in a subclass changes, we overload the method. Re-use this approach.
Composition is more flexible approach. OOP, from the other hand, helps to encapsulate the logic in a way that every class handles its own properties. So, for e.g. Humans, they add only info on hands on top of presentation prepared by the base class.
Please refactor in OOP-compliant manner

OleksiyRudenko: > This feature is quite flexible, but it can be added functionality for those inhabitants, who do not make sounds at all, such as fish. Мaybe gender for hermaphrodites.

We design classes in a manner that the code is easily extensible, yet we do not make pre-mature optimizations. SImply, because we cannot really predict what would be required.

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+  getFullName() {
+    return this.name
+  }
+}
+
+class Dog extends Inhabitant {
+  constructor(name, gender, legs) {
+    super('dog', name, gender, 'Woof-woof!')
+    this.legs = legs
```
OleksiyRudenko, line 32: Seems like all inhabitants have legs. Why handling them differently from name or gender?

OleksiyRudenko (changes requested):
@crealime let's make code OOP compliant.

submissions/crealime/oop-exercise/index.js
crealime: Fixed

crealime:
I've made changes. But I have questions.

Would it be correct to add more child classes to separate the inhabitants by gender? This way we could shorten the conditions in the methods. Or it would be redundant?

Could I create a class like "Speaking" and move the function "getFullInfo" into it, according to the OOP style? Because we can have many similar functions and the Inhabitants classes will grow. Or it is ok?

OleksiyRudenko:
> Would it be correct to add more child classes to separate the inhabitants by gender? This way we could shorten the conditions in the methods. Or it would be redundant?
Legal approach, but a bit of overkill. You could alternative streamline creating males and females in bulk.
E.g.
```
inhabitants = [
...["Dave", "Peter"].map(name => new Human(name, "male"),
...["Martha", "Anna"].map(name => new Human(name, "female"),
]
```
You could even have object literals instead of just names, e.g. `[{ name: "Dave", saying: "Heya!"}, { name: "Peter", saying: "Ciao!"}].map(...)`
Less repetition, declarative approach.

There is more than one legal solutions. Always a question of tradeoffs and specific use case.
Will it make code more clean? What are the code overheads? Performance?
 
> Could I create a class like "Speaking" and move the function "getFullInfo" into it, according to the OOP style? Because we can have many similar functions and the Inhabitants classes will grow. Or it is ok?

Again, it depends. Can this be solved with a simpler approach?
E.g.
```
const pronouns = {
  male: 'His',
  female: 'Her',
}

`${pronouns[this.gender]} name is...`
```

crealime:
Thank you! I understand. I will take this into account in the future. Here it is more convenient for me to work with named objects. I`ve added objects to define pronouns.

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,197 @@
...
+  getAcquaintanceInfo() {
+    return super.getAcquaintanceInfo().replace(this.name, this.name + ' ' + this.surname)
+  }
+
+  getLimbsInfo() {
+    return `${pronounsHeShe[this.gender]} have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
+  }
+
+  sayFullInfo() {
+    return `${this.saying} My name is ${this.name} ${this.surname} and I am a ${this.species}. My gender is ${this.gender}. I have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}. I have ${this.friends.length} ${this.friends.length === 1 ? 'friend' : 'friends'}${this.friends.length > 0 ? ': ' + this.friends.map(friend => friend.name).join(', ') : ''}.`
```
OleksiyRudenko, line 96: Looks quite heavy to read. As a peer developer I would refrain to take a task that would require to introduce changes to this code (adding a property, changing property handling).
Code is written for other people. Take care of others. You will want other developers take care of you.

You may want to introduce a method that would return a plural form of word along with its value. This would make the code shorter and more readable. Not sure if it helped to make it more maintainable and scalable.

submissions/crealime/oop-exercise/index.js
```diff
@@ -0,0 +1,197 @@
...
+    this.surname = surname
+    this.hands = hands
+  }
+
+  getAcquaintanceInfo() {
+    return super.getAcquaintanceInfo().replace(this.name, this.name + ' ' + this.surname)
+  }
+
+  getLimbsInfo() {
+    return `${pronounsHeShe[this.gender]} have ${this.hands || 0} ${this.hands === 1 ? 'hand' : 'hands'} and ${this.legs || 0} ${this.legs === 1 ? 'leg' : 'legs'}.`
```
OleksiyRudenko, line 92: So, if we need to change how output for legs looks like we'd need to introduce changes in two places.
Also it is a bad practice to directly refer to properties a class doesn't own.

OleksiyRudenko (approved):
@crealime quite a decent job.
Must admit I would feel comfortable to introduce any changes to the code before it is refactored and made more readable.

PR #344, opened 2022-08-29
Memory Pair Game

zhenyakornilov:
### Memory Pair Game

[Demo](https://zhenyakornilov.github.io/kottans-mpg/) |
[Code base](https://github.com/zhenyakornilov/kottans-mpg)

Please, review.

### DIFF
--- a/submissions/zhenyakornilov/memory-pair-game/js/index.js
+++ b/submissions/zhenyakornilov/memory-pair-game/js/index.js
@@ -1,0 +1,99 @@
+import cards from "./cards.js";
+
+const cardsMenu = document.getElementById("game-cards");
+const totalTriesCounter = document.getElementById("total-flips");
+const states = {
+  cardsMatches: 0,
+  firstFlippedCard: null,
+  secondFlippedCard: null,
+  currentCardPair: [],
+  totalTries: 0,
+};
+const allGameCards = shuffleCards([...cards, ...cards]);
+renderCards(allGameCards);
+
+function shuffleCards(cardsArr) {
+  for (let i = cardsArr.length - 1; i > 0; i--) {
+    let j = Math.floor(Math.random() * (i + 1));
+    let swap = cardsArr[i];
+    cardsArr[i] = cardsArr[j];
+    cardsArr[j] = swap;
+  }
+
+  return cardsArr;
+}
+
+function renderCards(cardsObj) {
+  const cardsAll = document.getElementById("game-cards");
+  let cardsHTML = cardsObj
+    .map(({ name, imageSrc }) => {
+      return `
+      <li class="game-card">
+        <div class="card-inner">
+          <div class="card-front">
+            <img class="card-img-front" src="images/treble_clef.webp" alt="image of treble clef">
+          </div>
+          <div class="card-back">
+            <img class="card-img" src="${imageSrc}" alt="image of ${name}">
+            <p class="card-desc">${name}</p>
+          </div>
+        </div>
+      </li>
+      `;
+    })
+    .join("");
+
+  cardsAll.innerHTML = cardsHTML;
+}
+
+function sleep(milliseconds) {
+  return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
+
+function flipBack() {
+  document.querySelectorAll(".flipped").forEach((card) => {
+    if (!card.classList.contains("matched")) {
+      card.closest(".game-card").classList.remove("flipped");
+    }
+  });
+}
+
+function restartGame() {
+  location.reload();
+}
+
+cardsMenu.addEventListener("click", async function flipCards({ target }) {
+  let currentCard = target.closest(".game-card");
+  if (currentCard) {
+    if (!currentCard.classList.contains("flipped")) {
+      states.currentCardPair.push(currentCard);
+    }
+    currentCard.classList.add("flipped");
+
+    if (states.currentCardPair.length === 2) {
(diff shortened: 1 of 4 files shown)


zhenyakornilov:
Added some changes with fixes and improvements

zhenyakornilov:
Feedback collected, issues fixed accordingly to remarks by other students. 

- Check [here](https://t.me/c/1382428271/51403)

zhenyakornilov:
Self checks done, please, remove stale label.

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+const cardsMenu = document.getElementById("game-cards");
+const totalTriesCounter = document.getElementById("total-flips");
+const states = {
+  cardsMatches: 0,
+  firstFlippedCard: null,
+  secondFlippedCard: null,
+  currentCardPair: [],
+  totalTries: 0,
+};
+const allGameCards = shuffleCards([...cards, ...cards]);
```
kasionio, line 12: Seems allCards or shuffledCards will be enough here, to make it consistent with shuffleCards

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+renderCards(allGameCards, cardsMenu);
+
+function shuffleCards(cardsArr) {
+  for (let i = cardsArr.length - 1; i > 0; i--) {
```
kasionio, line 16: Looks `sort` array method is the best choice here

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+      </li>
+      `;
+    })
+    .join("");
+
+  cardsList.innerHTML = cardsHTML;
+}
+
+function sleep(milliseconds) {
+  return new Promise((resolve) => setTimeout(resolve, milliseconds));
```
kasionio, line 49: why do we need async operation here?

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+        </div>
+      </li>
+      `;
+    })
+    .join("");
+
+  cardsList.innerHTML = cardsHTML;
+}
+
+function sleep(milliseconds) {
```
kasionio, line 48: I guess cards won't sleep, please try rename a function :)

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+function flipBack(currentCardPair) {
+  currentCardPair.forEach((card) => {
+    if (!card.classList.contains("matched")) {
+      card.closest(".game-card").classList.remove("flipped");
+    }
+  });
+}
+
+function checkCardsMatch() {
+  return states.firstFlippedCard.isEqualNode(states.secondFlippedCard);
```
kasionio, line 61: why so a complicated approach? I guess better compare name of the cards, even for performance reasons

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function restartGame() {
+  location.reload();
```
kasionio, line 65: clumsy method :))

submissions/zhenyakornilov/memory-pair-game/js/index.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+let cardsBlocked = false;
+
+cardsMenu.addEventListener("click", async function flipCards({ target }) {
```
kasionio, line 70: please explain why do you need async here

kasionio (changes requested):
Good job! Please take a look at the comments. 
Small tip - if we want to compare two elements we can simply use 
`function isElementsMatched (first, second) { return first.name === second.name }`

submissions/zhenyakornilov/memory-pair-game/js/index.js
zhenyakornilov: Did you mean that Fisher–Yates shuffle is overkill to shuffle game cards? :)

kasionio: We have a built-in `sort` method in JS for this purpose, no need to reinvent the wheel, also it is much better for code readability.

kasionio: But you can leave it, I changed my mind :))

submissions/zhenyakornilov/memory-pair-game/js/index.js
zhenyakornilov: I'm not sure I unterstood remark properly. Without async-await it all breaks and does not work as expected.

submissions/zhenyakornilov/memory-pair-game/js/index.js
zhenyakornilov: I wanted to create some kind of custom delay function instead of repeating **setTimeout** all the time. This is solution I came up with and thats why I use async - await later. I haven't figured out how to do that in another way yet (except using setTimeout in every place where I currently used my **sleep()** function.

zhenyakornilov:
I tried to account all the recommendations and remarks, but some questions still left.

PR #351, opened 2022-08-29
Task "Js pre oop"

DmitryHniezdilov:
### A Tiny JS World

[Demo](https://dmitryhniezdilov.github.io/task-a-tiny-JS-world/) |
[Code base](https://github.com/DmitryHniezdilov/task-a-tiny-JS-world/tree/gh-pages)

Please, review.

### DIFF
--- a/submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
+++ b/submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
@@ -34,7 +34,7 @@ const bear = {
   saying: "Think, think, think!",
 };
 
-const men = {
+const man = {
   species: "human",
   name: "Artur",
   gender: "male",
@@ -43,7 +43,7 @@ const men = {
   saying: "Do not borrow tomorrow's troubles today.",
 };
 
-const women = {
+const woman = {
   species: "human",
   name: "Marilyn",
   gender: "female",
@@ -52,12 +52,12 @@ const women = {
   saying: "A career is born in public – talent in privacy.",
 };
 
-const listOfInhabitants = [dog, cat, bear, men, women];
+const listOfInhabitants = [dog, cat, bear, man, woman];
 
 // ======== OUTPUT ========
 
-listOfInhabitants.forEach(
-  ({ species, name, gender, legs, hands, saying } = item) => {
-    print(`${species}; ${name}; ${gender}; ${legs}; ${hands}; ${saying}`);
-  }
-);
+listOfInhabitants.forEach((item) => {
+  const separator = "; ";
+  const listOfValues = Object.values(item).join(separator);
+  print(listOfValues);
+});
--- a/submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
+++ b/submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
@@ -1,0 +1,63 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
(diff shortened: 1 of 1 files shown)


submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,63 @@
...
+  saying: "A career is born in public – talent in privacy.",
+};
+
+const listOfInhabitants = [dog, cat, bear, men, women];
+
+// ======== OUTPUT ========
+
+listOfInhabitants.forEach(
+  ({ species, name, gender, legs, hands, saying } = item) => {
+    print(`${species}; ${name}; ${gender}; ${legs}; ${hands}; ${saying}`);
```
artem-trubin, line 61: There is a pattern here. A repeated action. Imagine a situation where you would need to change `;` with some other separator, it wouldn't be quick here if there would be more properties. Find a way to specify properties to print and then automate the output.

submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,63 @@
...
+const bear = {
+  species: "character",
+  name: "Winnie",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+  saying: "Think, think, think!",
+};
+
+const men = {
```
artem-trubin, line 37: `men` and `women` are plurals. Your objects describe a single man and a single woman.

artem-trubin (changes requested):
Hi, nice work, let's improve it a little.

submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
DmitryHniezdilov: THX for the correction. I'm confused 😕

submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
DmitryHniezdilov: THX for the explanation. I made the same automated output. Done

submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,63 @@
...
+  saying: "A career is born in public – talent in privacy.",
+};
+
+const listOfInhabitants = [dog, cat, bear, man, woman];
+
+// ======== OUTPUT ========
+
+listOfInhabitants.forEach((item) => {
+  const separator = "; ";
+  const listOfValues = Object.values(item).join(separator);
```
artem-trubin, line 61: Okay you're almost there. Task specifies not to use methods like `Object.values` and similar to them. Using them, you cannot rely on properties' order and presence. You can specify properties by hand, though.

artem-trubin (changes requested):

submissions/DmitryHniezdilov/task-a-tiny-JS-world/index.js
DmitryHniezdilov: THX for the correction. I hope that it's ok.

DmitryHniezdilov: This code I updated only in the work branch in the current repo :
https://github.com/DmitryHniezdilov/task-a-tiny-JS-world/blob/task-a-tiny-JS-world/index.js.

artem-trubin (approved):
Nice idea! Clever use of reduce.

DmitryHniezdilov:
Thank you.

PR #354, opened 2022-08-29
initial

natalia-artsiukh:
### OOP Exercise

[Demo](https://natalia-artsiukh.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/natalia-artsiukh/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/nataliaArtsiukh/OOP Exercise/index.js
+++ b/submissions/nataliaArtsiukh/OOP Exercise/index.js
@@ -1,0 +1,75 @@
+class Habitant {
+  constructor(name, gender, saying, friends, legs) {
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+    this.legs = legs;
+  }
+}
+
+class Human extends Habitant {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super (name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+  toString() {
+    let {name, gender, saying, friends, legs, hands, constructor} = this;
+    return `${saying}! My name is ${name}. I'm ${gender}, ${constructor.name.toLowerCase()}. I have ${hands} hands and ${legs} legs. I'm friends with ${friends}.`;
+  }
+}
+
+class Cat extends Habitant {
+  constructor(name, gender, saying, friends, legs = 4) {
+    super(name, gender, saying, friends, legs);
+  }
+  scatterLitterBox() {
+    let leftoverLitter = Math.random();
+    if (leftoverLitter < 0.5) {
+      return 'cat litter is everywhere. Time to clean.'
+    } else {
+      return 'around the tray is clean. Check if the cat is alive.'
+    }
+  }
+  toString() {
+    let {name, gender, saying, friends, legs, constructor} = this;
+    return `${saying}! This cat's name is ${name}. It's ${gender}, ${constructor.name.toLowerCase()}. It has ${legs} legs. It's friends with ${friends}. By the way, ${this.scatterLitterBox()}`;
+  }
+}
+
+class Dog extends Habitant {
+  constructor(name, gender, saying, friends, company, legs = 4) {
+    super(name, gender, saying, friends, legs);
+    this.company = company;
+  }
+  sniff() {
+    let index = Math.floor(Math.random() * this.company.length);
+    if (this.company[index] instanceof Cat) {
+      return 'Cat is among us. Woof-woof!'
+    } else {
+      return `No cat around. ${this.name} wags its tail.`
+    }
+  }
+  toString() {
+    let {name, gender, saying, friends, legs, constructor} = this;
+    return `${saying}! This dog's name is ${name}. It's ${gender}, ${constructor.name.toLowerCase()}. It has ${legs} legs. It's friends with ${friends}. ${name} is sniffing. ${this.sniff()}`;
+  }
+}
+
+class CatWoman extends Cat {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super(name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+}
+
+CatWoman.prototype.toString = Human.prototype.toString;
+
+const habitants = [];
+habitants.push(new Human('Bob', 'male', 'I\'m vegan', 'no one'));
+habitants.push(new Human('Mary', 'female', 'Hi everyone', 'Tom, Rex, Halle'));
+habitants.push(new Cat('Tom', 'male', 'Mrrr', 'Mary, Halle'));
+habitants.push(new Dog('Rex', 'male', 'Woof', 'Mary', habitants));
+habitants.push(new CatWoman('Halle', 'female', 'Mrrr', 'Mary, Tom, Batman'));
+ 
+habitants.forEach((habitant) => print(habitant));


natalia-artsiukh:
When I did the work, I checked this list. In my opinion everything is ok

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+CatWoman.prototype.toString = Human.prototype.toString;
+
+const habitants = [];
+habitants.push(new Human('Bob', 'male', 'I\'m vegan', 'no one'));
+habitants.push(new Human('Mary', 'female', 'Hi everyone', 'Tom, Rex, Halle'));
+habitants.push(new Cat('Tom', 'male', 'Mrrr', 'Mary, Halle'));
+habitants.push(new Dog('Rex', 'male', 'Woof', 'Mary', habitants));
+habitants.push(new CatWoman('Halle', 'female', 'Mrrr', 'Mary, Tom, Batman'));
```
OleksiyRudenko, line 73: Too much repetition of `habitants.push`...
Check the following example:
```
class A {
  constructor(v) { this.v = v }
}

[
  {
    cl: A,
    v: ['abc', 567]
  },
  {
    cl: A,
    v: ['def',879],
  },
].map(({cl, v}) => new cl(...v))
```

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+ 
+habitants.forEach((habitant) => print(habitant));
```
OleksiyRudenko, line 75: Once you have `Array#map` above you will be able to chain `Array#forEach`

OleksiyRudenko (changes requested):
@natalia-artsiukh very well done!
A few improtant improvements need tho.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+class Human extends Habitant {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super (name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+  toString() {
+    let {name, gender, saying, friends, legs, hands, constructor} = this;
+    return `${saying}! My name is ${name}. I'm ${gender}, ${constructor.name.toLowerCase()}. I have ${hands} hands and ${legs} legs. I'm friends with ${friends}.`;
+  }
```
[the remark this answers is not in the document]
natalia-artsiukh: Is it about toLowerCase? Or something else? I don't understand(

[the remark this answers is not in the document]
OleksiyRudenko: By default a property defined in a class should be handled by defining class only:
- Parent class should never refer to properties of any child classes as it restricts child class' "freedom" and "independency".
- There should be good reasons to directly refer properties of a base class from a child class. There are no such reasons in this task. An non-specific example of a good reason: e.g. multiply the value of an inherited property in child class as it makes sense.

Reference: Open-closed principle from SOLID, encapsualtion from OOP.

So how do we "extract" data from parent class in this exercise?
A child class may call parent's method that handles its own, parent's, properties, and uses whatever that method returns.

How do we avoid having different method names for the basically same job thus unnecessarily cluttering the namespace?
We benefit from method overloading (exactly similar method name across all classes) and specific `super` ability.
Check materials re `super` in TJSW project docs.

submissions/nataliaArtsiukh/OOP Exercise/index.js
OleksiyRudenko: This way we restrict ourselves with what is already exist. Hence no "circular" or mutual friends possible. And even so are forced to set friends in a specific order.

But we have freedom to call any object method and modify object's properties after it has been created.

submissions/nataliaArtsiukh/OOP Exercise/index.js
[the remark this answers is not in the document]
natalia-artsiukh: Now the properties of the parent are processed in the parent (if I understood correctly what was required of me). But in order to keep the line beautiful, while not plagiarizing the version with an object with separate words/phrases (as you showed yesterday on the stream), I had to additionally split the result. I haven't found a more elegant solution.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class Human extends Habitant {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super (name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+  toString() {
+    let {name, gender, saying, friends, legs, hands, constructor} = this;
```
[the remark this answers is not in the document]
natalia-artsiukh: Done

submissions/nataliaArtsiukh/OOP Exercise/index.js
natalia-artsiukh: Done

submissions/nataliaArtsiukh/OOP Exercise/index.js
natalia-artsiukh: I find the Dog in the finished array and give it a link to all the habitants. This must be done before printing, so it is not possible to do chaining.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,100 @@
+class Habitant {
+  constructor(name, gender, saying, friends, legs) {
+    this.name = name;
+    this.gender = gender;
+    this.saying = saying;
+    this.friends = friends;
+    this.legs = legs;
+  }
+  toString() {
+    let {name, gender, saying, friends, legs} = this;
```
OleksiyRudenko, line 10: `const` is more approriate here. `let` implies (and creates expectation) we are going to change the value.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+class CatWoman extends Cat {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super(name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+}
+
+CatWoman.prototype.toString = Human.prototype.toString;
+
+let habitants = [
```
OleksiyRudenko, line 75: Same as above re `const` vs `let`.
Also `const` doesn't restrict us in mutating the associated array. Key here that we are not going to assign a different array to the same variable.
Altogether, there are very few use cases for `let`.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+  {
+    habitantClass: Human,
+    habitantProperties: ['Bob', 'male', 'I\'m vegan', 'no one']
```
OleksiyRudenko, line 78: Containing variable name provides context. So we may skip `habitant` in property names.
Although `class` is a reserved word so you cannot name a property or a variable. Same refers to other keywords (e.g. `return`).
The convention is to use `className` instead.

A workaround like `"class"` would work, however is a bad practice.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+    if (this.company[index] instanceof Cat) {
+      return 'Cat is among us. Woof-woof!'
+    } else {
+      return `No cat around. ${this.name} wags its tail.`
+    }
+  }
+  toString() {
+    let {constructor: {name: species}} = this;
+    let [name, gender, saying, friends, legs] = super.toString().split(';');
+    return `${saying}! This dog's name is ${name}. It's ${gender}, ${species.toLowerCase()}. It has ${legs} legs. It's friends with ${friends}. ${name} is sniffing. ${this.sniff()}`;
```
OleksiyRudenko, line 62: Yeah, not a really elegant way to handle required order. But I can see that you are able to handle things as normally expected, so let this be.

submissions/nataliaArtsiukh/OOP Exercise/index.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+class CatWoman extends Cat {
+  constructor(name, gender, saying, friends, legs = 2, hands = 2) {
+    super(name, gender, saying, friends, legs);
+    this.hands = hands;
+  }
+}
```
OleksiyRudenko, line 71: To make long story short:
OOP is not good for hybrids, and it is not your fault. Composition pattern is to the rescue but it is a completely different approach to the design of system of domain objects. Just keep in mind.

OleksiyRudenko (approved):
@natalia-artsiukh great job!
Check the comments below for a bit of learnings.

PR #375, opened 2022-08-31
Tiny js world

VladimirRutskiy:
### OOP exercise

[Demo](https://vladimirrutskiy.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/VladimirRutskiy/a-tiny-JS-world)

Please, review.


VladimirRutskiy:
i have completed self-checks

OleksiyRudenko:
@VladimirRutskiy 
I am confused: is it code for `TJSW` or for `OOP Exercise`?

VladimirRutskiy:
@OleksiyRudenko  It's code for "10. OOP exercise - practice from Advanced Topics list " task

artem-trubin (changes requested):
Hi! I don't see any code that should create instances of your classes or print their info. It seems that this file is incomplete. Can you fix this, please?

VladimirRutskiy:
@artem-trubin I'm so sorry. I didn't notice that the file was partly fill

submissions/VolodymyrRutskyi/tiny-js-world/index.js
```diff
@@ -1,90 +1,52 @@
...
+      this.saying = saying;
+      this.props = ['species', 'name', 'gender', 'legs', 'saying']
+   }
+   showProps() {
+      return this.props.map(prop => this[prop]).join('; ');
+   }
+}
+
+class Human extends Inhabitant{
+   constructor (species, name, gender, saying){
```
artem-trubin, line 16: Are you sure a `Human` needs a `species` here? A human can only be human, isn't it?

submissions/VolodymyrRutskyi/tiny-js-world/index.js
```diff
@@ -1,90 +1,52 @@
...
+      super(species, name, gender, 2, saying);
+      this.hands = 2;
+      this.props = [...this.props, 'hands']
+   }
+}
+class Woman extends Human{
+   constructor(name, gender, saying){
+      super('woman', name, gender, saying)
```
artem-trubin, line 24: A woman can only be female, so there's no need to ask for a gender in the constructor here.

artem-trubin (changes requested):
Hey, nice work. Let's make it better. See the comments below.

VladimirRutskiy:
@artem-trubin Thank you. I fixed all remarks

artem-trubin (approved):
Good job!

submissions/VolodymyrRutskyi/tiny-js-world/index.js
```diff
@@ -1,90 +1,52 @@
...
+   showProps() {
+      return this.props.map(prop => this[prop]).join('; ');
+   }
+}
+
+class Human extends Inhabitant{
+   constructor (name, gender, saying){
+      super('human', name, gender, 2, saying);
+      this.hands = 2;
+      this.props = [...this.props, 'hands']
```
OleksiyRudenko, line 19: It is generally a bad practice to directly change properties that are owned by the base class.
Every class should handle its own properties i.e. encapsulate both properties and behaviour.
The same refers to `Inhabitant#showProps`.
Let every class handle properties it owns.
Use method overloading to provide seamless user experience.
You will benefit from explicit call to inherited version of the methos via `super`.
Task and original project materials offer links to the articles that you will find helpful.

OleksiyRudenko (changes requested):
@VladimirRutskiy very well done!
Although need to improve it for the sake of OOP compliance, which in its turn will make code maintainable and extensible, concerns will be separated across classes.

submissions/VolodymyrRutskyi/tiny-js-world/index.js
VladimirRutskiy: I have fixed

OleksiyRudenko (changes requested):
@VladimirRutskiy make a step back.
`legs` were right where they belonged.
My comment referred to `this.prop` direct assignment in a child class.
Actually, you do not need `this.props` at all. It is brought from pre-OOP version obviously, and there it was a workaround. OOP paradigm helps to avoid that sort of workarounds.
Let's restore previous version. I will explain some theory on that and we will rather merge that previous version. It was better.

artem-trubin (approved):
Good job!

PR #376, opened 2022-08-31
a tiny js world

drewzag:
### A Tiny JS World

[Demo](https://drewzag.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/drewzag/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/drewzag/a-tiny-js-world/index.js
+++ b/submissions/drewzag/a-tiny-js-world/index.js
@@ -1,0 +1,61 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/drewzag/a-tiny-JS-world
+   Web app: https://drewzag.github.io/a-tiny-JS-world/
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+
+const dog = {
+  species: 'dog',
+  name: 'Jack',
+  gender: 'male',
+  legs: 4,
+  hands: 0,
+  saying: 'bark-bark!',
+}
+
+const cat = {
+  species: 'cat',
+  name: 'Tom',
+  gender: 'male',
+  legs: 4,
+  hands: 0,
+  saying: 'meow',
+}
+
+const woman = {
+  species: 'human',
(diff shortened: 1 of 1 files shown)


submissions/drewzag/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,61 @@
...
+  legs: 2,
+  hands: 2,
+  saying: cat.saying,
+}
+
+// ======== OUTPUT ========
+
+const inhabitants = [dog, cat, woman, man, catWoman]
+
+const printInhabitants = (inhabitants) => Object.values(inhabitants).join('; ')
```
artem-trubin, line 59: Doesn't this function receive 1 inhabitant?

artem-trubin, line 59: The task specifically states not to use methods like `Object.values` and other similar ones.

submissions/drewzag/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,61 @@
...
+  saying: 'Dobriy den, everybody!',
+}
+
+const catWoman = {
+  species: 'cat-human',
+  name: 'Cat Marry',
+  gender: 'female',
+  legs: 2,
+  hands: 2,
+  saying: cat.saying,
```
artem-trubin, line 52: If we change `cat.saying`, would `catWoman.saying` change too? The task requires it to do so.

artem-trubin (changes requested):
Hey, nice work, let's improve it. Check the comments below.

submissions/drewzag/a-tiny-js-world/index.js
drewzag: honestly, I think yes, it would change)

drewzag:
@artem-trubin thanks for review)

submissions/drewzag/a-tiny-js-world/index.js
drewzag: Oh, sorry) I totally forget about getter and setter

artem-trubin (approved):
That's one way of doing this, good job! You could've also used `bind` method or connected them with prototyping.

PR #381, opened 2022-08-31
Add Tiny JS World task

AriadnaKravchuk:
### A Tiny JS World

[Demo](https://ariadnakravchuk.github.io/a-tiny-JS-world) |
[Code base](https://github.com/AriadnaKravchuk/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/AriadnaKravchuk/tiny-js-world/index.js
+++ b/submissions/AriadnaKravchuk/tiny-js-world/index.js
@@ -59,15 +59,17 @@ const catWoman = {
 
 // ======== OUTPUT ========
 
-function printInhabitant(inhabitant) {
-   let result = "";
-
-   for (let val in inhabitant) {
-      result += inhabitant[val] + "; ";
-   }
-
-   print(result);
-}
-
 const inhabitants = [dog, cat, man, woman, catWoman];
-inhabitants.forEach(printInhabitant);
+const keys = [
+   "species",
+   "name",
+   "gender",
+   "legs",
+   "hands",
+   "saying",
+   "friends",
+];
+
+inhabitants.forEach((inhabitant) =>
+   print(keys.map((key) => inhabitant[key]).join("; "))
+);
--- a/submissions/AriadnaKravchuk/tiny-js-world/index.js
+++ b/submissions/AriadnaKravchuk/tiny-js-world/index.js
@@ -14,7 +14,7 @@ const dog = {
    legs: 4,
    hands: 0,
    saying: "woof-woof!",
-   friends: "Cake",
+   friends: "Cake"
 };
 
 const cat = {
@@ -24,7 +24,7 @@ const cat = {
    legs: 4,
    hands: 0,
    saying: "meow!",
-   friends: ["Pie", "Selina"],
+   friends: ["Pie", "Selina"]
 };
 
 const man = {
@@ -34,7 +34,7 @@ const man = {
    legs: 2,
    hands: 2,
    saying: "Hello!",
-   friends: "Dominica",
+   friends: "Dominica"
 };
 
 const woman = {
@@ -44,7 +44,7 @@ const woman = {
    legs: 2,
    hands: 1,
    saying: "Hello!",
-   friends: ["Dominic", "Selina"],
+   friends: ["Dominic", "Selina"]
 };
 
 const catWoman = {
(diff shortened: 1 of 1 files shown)


AriadnaKravchuk:
I completed self-checks

submissions/AriadnaKravchuk/tiny-js-world/index.js
```diff
@@ -0,0 +1,75 @@
...
+   friends: ["Dominic", "Selina"]
+};
+
+const catWoman = {
+   species: "human",
+   name: "Selina",
+   gender: "female",
+   legs: 2,
+   hands: 2,
+   saying: cat.saying,
```
artem-trubin, line 56: If we change `cay.saying`, would `catWoman.saying` change as well? The task requires catwoman to always say what the cat does.

artem-trubin (changes requested):
Hey, nice work! Let's add an improvement, check the comment below.

artem-trubin:
It works now, good job!

AriadnaKravchuk:
> It works now, good job!

Thanks, and I have a question. What constants should be written in uppercase? (or I need use uppercase for all constant).

artem-trubin:
> > It works now, good job!
> 
> Thanks, and I have a question. What constants should be written in uppercase? (or I need use uppercase for all constants).

## ALL UPPERCASE. LIKE THIS.

In all uppercase are usually those constants that are like "settings" if you know what I mean. Those that you write globally in the start of the file, like "MAIN_BUTTON_COLOR". Something that will 100% never change in the program and if it should be changed, you would go to this constant and change it there by hand. Or, sometimes, if you want to have some kind of string-based state, like a status or something. Imagine a flight as an object and you want to add a property `flight.status` that can be one of some different string statuses, like `in-progress`, `landed`, `setting-off` or something like this. To prevent yourself from making a simple mistake in these strings further in code, these could be defined as global constants like this:
```js
const STATUS_IN_PROGRESS = 'in-progress'
const STATUS_LANDED = 'landed'
const STATUS_SETTING_OFF = 'setting-off`

if (flight.status === STATUS_IN_PROGRESS) { // ...
```

This gives you several advantages:
1. When you type `STATUS_`, your code editor probably hands you all statuses as a hint, since they're in the global scope.
2. Protects you from typos on the building stage, since you'll just get an error if you write a non-existant variable name.

## FirstLetterUppercase. LikeThis.
This style is usually used to describe a class or a constructor function. Something that can later spawn instances of itself. Like **I**nhrabitant can spawn **i**nhabitants.

I hope I was descriptive enough, feel free to ask more questions. Here's a cute possum for your curiosity:

image

AriadnaKravchuk:
> > > It works now, good job!
> > 
> > 
> > Thanks, and I have a question. What constants should be written in uppercase? (or I need use uppercase for all constants).
> 
> ## ALL UPPERCASE. LIKE THIS.
> In all uppercase are usually those constants that are like "settings" if you know what I mean. Those that you write globally in the start of the file, like "MAIN_BUTTON_COLOR". Something that will 100% never change in the program and if it should be changed, you would go to this constant and change it there by hand. Or, sometimes, if you want to have some kind of string-based state, like a status or something. Imagine a flight as an object and you want to add a property `flight.status` that can be one of some different string statuses, like `in-progress`, `landed`, `setting-off` or something like this. To prevent yourself from making a simple mistake in these strings further in code, these could be defined as global constants like this:
> 
> ```js
> const STATUS_IN_PROGRESS = 'in-progress'
> const STATUS_LANDED = 'landed'
> const STATUS_SETTING_OFF = 'setting-off'
> 
> if (flight.status === STATUS_IN_PROGRESS) { // ...
> ```
> 
> In fact, you're still operating string values, but in code - they're wrapped in constants.
> 
> This gives you several advantages:
> 
> 1. When you type `STATUS_`, your code editor probably hands you all statuses as a hint, since they're in the global scope.
> 2. Protects you from typos on the building stage, since you'll just get an error if you write a non-existant variable name.
> 
> ## FirstLetterUppercase. LikeThis.
> This style is usually used to describe a class or a constructor function. Something that can later spawn instances of itself. Like **I**nhrabitant can spawn **i**nhabitants.
> 
> I hope I was descriptive enough, feel free to ask more questions. Here's a cute possum for your curiosity:
> 
> image

So, object like cat, dog etc should be written in FirstLetterUppercase?

artem-trubin:
No, regular objects are lowercase. If it's a class or a constructor function (in other words, something that can create instances) - then we use FirstLetterUppercase.

AriadnaKravchuk:
> No, regular objects are lowercase. If it's a class or a constructor function (in other words, something that can create instances) - then we use FirstLetterUppercase.

Okay, thank you!

PR #384, opened 2022-09-01
Memory pair game

Tedzury:
### Memory pair game

[Demo](https://tedzury.github.io/Memory_pair_game/) |
[Code base](https://github.com/Tedzury/Memory_pair_game)

Please, review.

### DIFF
--- a/submissions/Oleksii_Drohachov/Memory_pair_game/script.js
+++ b/submissions/Oleksii_Drohachov/Memory_pair_game/script.js
@@ -1,0 +1,244 @@
+const mainContainer = document.querySelector('.container')
+const modal = document.querySelector('.modal__container');
+const modalBtn = document.querySelector('.modal__btn');
+const gameboard = document.querySelector('.gameboard');
+const scoreCount = document.querySelector('.score__container');
+
+
+
+renderModal('You are welcome, my friend ! Wanna play some memory game to endure your mind and become more powerfull ? Try to find pairs of cards as fast as you can! Tap the button to begin... ', 'START THE GAME')
+
+function renderModal(titleText, btnText) {
+
+    const modalWrapper = document.createElement('div');
+    modalWrapper.classList.add('modal__wrapper');
+    const modalTitle = document.createElement('p');
+    modalTitle.classList.add('modal__title');
+    modalTitle.textContent = titleText;
+    const modalBtn = document.createElement('button');
+    modalBtn.classList.add('modal__btn');
+    modalBtn.textContent = btnText;
+    modalBtn.addEventListener('click', startGame)
+    const closeBtn = document.createElement('button');
+    closeBtn.classList.add('modal__btn');
+    closeBtn.textContent = 'NO, THANKS';
+    closeBtn.addEventListener('click', hideModal);
+
+    modalWrapper.append(modalTitle);
+    modalWrapper.append(modalBtn);
+    modalWrapper.append(closeBtn);
+
+    modal.append(modalWrapper);
+    modal.classList.add('shown');
+}
+
+function renderModalForReset() {
+    renderModal('Are you sure, you want to restart the game?', 'RESET')
+}
+
+function renderModalAfterWin() {
+    renderModal('Good job my friend! You became even stronger! Want to start one more time?', 'RESET')
+}
+
+function hideModal() {
+    modal.classList.remove('shown');
+    modal.innerHTML ='';
+}
+
+function startGame() {  
+    scoreCount.innerHTML = '';
+    gameboard.innerHTML = '';
+    modal.innerHTML = '';
+    modal.classList.remove('shown');
+    prepareGame();
+}
+
+function prepareGame() {
+
+    const cardsArray = [
+        {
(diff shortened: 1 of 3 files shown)


Tedzury:
Nooooo, bot, I'm not stale, I'm still here !

Tedzury:
I'm still here...

Tedzury:
UAT - done. Feel myself ready for review.

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,220 @@
...
+function prepareGame() {
+  const cardsArray = [
+    {
+      name: "beast",
+      path: "img/Beast.jpg",
+    },
+    {
+      name: "beast",
+      path: "img/Beast.jpg",
+    },
```
A-Ostrovnyy, line 68: No need to duplicate card items

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,220 @@
...
+    {
+      name: "metal",
+      path: "img/Metal.jpg",
+    },
+    {
+      name: "metal",
+      path: "img/Metal.jpg",
+    },
+  ];
+  let shuffled = [...cardsArray];
```
A-Ostrovnyy, line 110:
```suggestion
  let shuffled = [...cardsArray, ...cardsArray];
```
Also, if you don't reassign it use const instead.

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,220 @@
...
+      card.append(frontPage);
+      card.append(backPage);
+      cardWrapper.append(card);
+    });
+    gameboard.append(cardWrapper);
+  }
+
+  function cardsHandler() {
+    const cards = document.querySelectorAll(".card");
+    cards.forEach((card) => card.addEventListener("click", flipCard));
```
A-Ostrovnyy, line 163: User event delegation. [Read more](https://davidwalsh.name/event-delegation)

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,220 @@
...
+
+    let hasFlipped = false;
+    let lockBoard = false;
+    let firstCard, secondCard;
+
+    function flipCard() {
+      if (lockBoard) return;
+      if (this === firstCard) return;
+
+      this.classList.add("flip");
```
A-Ostrovnyy, line 173: You can easily lose context and the corresponding value of `this` pass selected card as a parameter for the function.

A-Ostrovnyy (changes requested):
Good work!

Tedzury:
Good day. Fixed everuthing you mentioned, applied Event delegation to gameboard, so slightly reefactored game logic, some tricks with event targets. Also refactored modal rendering, put out making of modal from function, in func changing only modal text and button title.

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,203 @@
...
+      card.append(backPage);
+      cardWrapper.append(card);
+    });
+    gameboard.append(cardWrapper);
+  }
+
+  function cardsHandler() {
+    gameboard.addEventListener("click", ({ target }) => {
+      if (
+        target.parentElement.classList.contains("card") &&
```
A-Ostrovnyy, line 132: Don't use constructions like children[0], firstElementChild, parentNode, nextSibling, etc. In such way, you rely on the order of DOM elements. So in case when you will have changed design - your code will be broken. Which is bad, obviously. Use querySelector or closest, if in event, instead. [Read more](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

submissions/Oleksii_Drohachov/Memory_pair_game/script.js
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function startGame() {
+  scoreCount.innerHTML = "";
+  gameboard.innerHTML = "";
+  hideModal();
+  prepareGame();
+}
+
+function prepareGame() {
```
A-Ostrovnyy, line 49: Your function calls `prepareGame` but inside this function, you describe all game logic.
Single responsibility - your functions should do only one job. As an example function in which you fetch users should only fetch them and not render, transform or process them in other ways. All these things should be done in another place, outside your function. The same applied to the sort function, which usually does all types of sorting 😉

A-Ostrovnyy (changes requested):
Good progress!

Tedzury:
Good day. Applied target.closest to define clicked card. Also refactored prepare game function, splited it to func, that renders gameboard, func, that handles card click and resposible for game logic, and separetely pulled out rendering of scoreboard. Now it's rendered only once and then changing only one number of score. Previosly scoreboard was deleted everytime game resets.

PR #388, opened 2022-09-01
DOM task

franchukv:
### DOM

[Demo](https://franchukv.github.io/dom-practice/) |
[Code base](https://github.com/franchukv/dom-practice)

Please, review.


franchukv:
Self-check done.

submissions/franchukv/dom-api/script.js
```diff
@@ -0,0 +1,137 @@
...
+    for (item in data) {
+        let navItem = document.createElement('li');
+        navItem.classList.add('cards__item');
+
+        let h3 = document.createElement('h3');
+        h3.classList.add('cards__title');
+        h3.textContent = item;
+        navItem.appendChild(h3);
+
+        navigation.append(navItem);
```
A-Ostrovnyy, line 107: Adding elements to DOM from a loop is a bad practice. A browser will run reflow and repaint for every element in the loop. Instead, you can:

1. Use append method, which can add several elements in one operation.
2. Create some wrapper, add your items to the wrapper and then add it to DOM. It will be one operation.
3. Clone current container. Add items to a container and then replace your old container with a new one. But be aware of event listeners.
4. Use innerHTML instead.

submissions/franchukv/dom-api/script.js
```diff
@@ -0,0 +1,137 @@
...
+
+    const title = document.createElement('h2');
+    title.classList.add('content-title');
+    title.textContent = item;
+    mainContent.append(title);
+
+    const text = document.createElement('div');
+    text.classList.add('content-text');
+    text.innerHTML = data[item].text;
+    mainContent.append(text);
```
A-Ostrovnyy, line 127: You can use a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to define the whole item markup instead of texting all these lines.

submissions/franchukv/dom-api/script.js
```diff
@@ -0,0 +1,137 @@
...
+
+    createContent(event.target.textContent);
+}
+
+function createNavigation(item) {
+    for (item in data) {
+        let navItem = document.createElement('li');
+        navItem.classList.add('cards__item');
+
+        let h3 = document.createElement('h3');
```
A-Ostrovnyy, line 102: In your navigation user clicks on the heading. The heading is not an interactive element, use the corresponding HTML tag for that.

A-Ostrovnyy (changes requested):
Good work!
It will be nice to highlight selected element in navigation.

franchukv:
Done all recommendations from latest review.

A-Ostrovnyy (approved):
Well done, for the future, don't resolve comments, it's for authors)

PR #390, opened 2022-09-01
a-tiny-js-world

kuslosolodka:
### A-Tiny-JS-World

[Demo](https://kuslosolodka.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/kuslosolodka/a-tiny-JS-world.git)

Please, review.

### DIFF
--- a/submissions/kuslosolodka/a-tiny-js-world/index.js
+++ b/submissions/kuslosolodka/a-tiny-js-world/index.js
@@ -47,7 +47,7 @@ const cat_woman = {
   hands: 2,
   saying: cat["saying"],
 };
-dog.friend = `${woman.name},${man.name},${woman.name}`;
+dog.friend = `${woman.name},${man.name},${cat.name}`;
 cat.friend = undefined;
 woman.friend = `${man.name},${dog.name}`;
 man.friend = `${woman.name},${dog.name}`;
--- a/submissions/kuslosolodka/a-tiny-js-world/index.js
+++ b/submissions/kuslosolodka/a-tiny-js-world/index.js
@@ -36,9 +36,9 @@ const cat_woman = {
   gender: "female",
   legs: 2,
   hands: 2,
-  saying: cat.saying,
 };
 
+Object.setPrototypeOf(cat_woman, cat);
 const inhabitants = [dog, cat, woman, man, cat_woman];
 const keys = ["species", "name", "gender", "legs", "hands", "saying", "friend"];
 
--- a/submissions/kuslosolodka/a-tiny-js-world/index.js
+++ b/submissions/kuslosolodka/a-tiny-js-world/index.js
@@ -1,18 +1,9 @@
-/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
-   Complete the below for code reviewers' convenience:
-
-   Code repository: _put repo URL here_
-   Web app: _put project's github pages URL here_
-   */
-
-// ======== OBJECTS DEFINITIONS ========
-// Define your objects here
 const dog = {
   species: "dog",
   name: "Brovko",
   gender: "male",
   legs: 4,
-  hands: undefined,
+  hands: 0,
   saying: "woof-woof!",
 };
 const cat = {
@@ -20,7 +11,7 @@ const cat = {
   name: "Murka",
   gender: "female",
   legs: 4,
-  hands: undefined,
+  hands: 0,
   saying: "meow!",
 };
 const woman = {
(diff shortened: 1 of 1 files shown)


submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+const cat_woman = {
+  species: "supernatural",
+  name: "Batgirl",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: cat["saying"],
+};
+dog.friend = `${woman.name},${man.name},${cat.name}`;
+cat.friend = undefined;
```
artem-trubin, line 51: `undefined` is a default value for some variable that was not given a value or for places where we wait for a value, but don't get it as a result of some mistake or something like that. If you want to specify something as "nothing" or "empty" use `null`. But of course, the end user of the app/website should see neither `null` nor `undefined`. You release your `undefined` to the eyes of the user, and that is a mistake. For the sake of this task, animals are allowed to have `0` hands, for example. And if someone doesn't have `friends` - you should just ignore that property.

submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+};
+const cat_woman = {
+  species: "supernatural",
+  name: "Batgirl",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: cat["saying"],
+};
+dog.friend = `${woman.name},${man.name},${cat.name}`;
```
artem-trubin, line 50: `friend` should be plural, I guess. And it should be an array since here is where you keep your "raw" data, a model if you will. Then some other code should pick this data and format it to display.

submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+  hands: 2,
+  saying: "I am Marko!",
+};
+const cat_woman = {
+  species: "supernatural",
+  name: "Batgirl",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: cat["saying"],
```
artem-trubin, line 48: Catwoman's saying should change if we change the cat's saying any time. Does yours?

submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+const inhabitants = [dog, cat, woman, man, cat_woman];
+inhabitants.map((item) => {
+  let res = "";
+  for (let key in item) {
```
artem-trubin, line 76: You should use `Array.forEach` and other similar methods everywhere where it's possible. It increases the readability of the code a lot. Also, don't rely on cycling through keys in the objects, you never know for sure which keys are present or in which order. Find a way to _specify_ the keys by hand and then get values from objects.

submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,80 @@
...
+   print('<strong>ABC</strong>');
+   print('<strong>ABC</strong>', 'div');
+
+   print('human; John; male; 2; 2; Hello world!; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny');
+   print('human; <strong>John</strong>; male; 2; 2; <em>Hello world!</em>; Rex, Tom, Jenny', 'div');
+   */
+const inhabitants = [dog, cat, woman, man, cat_woman];
+inhabitants.map((item) => {
+  let res = "";
```
artem-trubin, line 75: Instead of stacking values into the string like that, it's cleaner to prepare some values and then `.join()` them.

artem-trubin (changes requested):
Nice work! Let's improve it. Check the comments below.

submissions/kuslosolodka/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,50 @@
...
+  hands: 2,
+  saying: "I am Marko!",
+};
+const cat_woman = {
+  species: "supernatural",
+  name: "Batgirl",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: cat.saying,
```
artem-trubin, line 39: Nice work! Your catwoman's saying is still not connected to cat's saying. Yes, catwoman says what cat said *initially*, but if we change cat's saying later in code, catwoman will not change her saying accordingly.

artem-trubin (changes requested):

artem-trubin (approved):
Yes, this is one way of doing so. I'm going to approve that, but be aware that following [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf), the `Object.setPrototypeOf` is a very slow operation and you could use `Object.create` to achieve the same goal. You could've also just used `bind`. That's just for the future, look into those methods. Great work!

PR #399, opened 2022-09-02
OOP_exercise

Oleksandr-Heleta:
### OOP_exercise

[Demo](https://oleksandr-heleta.github.io/a-tiny-JS-world) |
[Code base](https://github.com/Oleksandr-Heleta/a-tiny-JS-world/tree/populate-world)

Please, review.

### DIFF
--- a/submissions/Oleksandr-Heleta/OOP_exercise/index.js
+++ b/submissions/Oleksandr-Heleta/OOP_exercise/index.js
@@ -14,15 +14,10 @@ class Inhabitants {
       this.gender = gender;
       this.legs = legs;
       this.say = say;
-      this.propertiesOrder = ['species', 'name', 'gender', 'legs', 'say']
-   }
-
-   addProperty(property, order) {
-      this.propertiesOrder.splice(order, 0, property);
    }
 
    createString() {
-      return this.propertiesOrder.map((property) => this[property] || 'none');
+      return `${this.species}, ${this.name}, ${this.gender}, ${this.legs}, ${this.say};`;
    }
 }
 
@@ -30,7 +25,10 @@ class Human extends Inhabitants {
    constructor(name, gender, say) {
       super('human', name, gender, 2, say);
       this.hands = 2;
-      this.addProperty('hands', 4);
+   }
+
+   createString() {
+      return `${this.species}, ${this.name}, ${this.gender}, ${this.legs}, ${this.hands}, ${this.say};`;
    }
 }
 
@@ -47,16 +45,20 @@ class Dog extends Animal {
 }
 
 class Cat extends Animal {
+   static say = 'moew';
    constructor(name, gender) {
-      super('cat', name, gender, "meow");
+      super('cat', name, gender, Cat.say);
+
+   }
+
+   static sayingPhrase(phrase) {
+      return Cat.say = phrase || 'meow';
    }
 }
 
 class CatWoman extends Human {
    constructor(name, gender) {
-      super(name, gender)
-      this.mutation = new Cat();
-      this.say = this.mutation.say
+      super(name, gender, Cat.say)
    }
 }
 
@@ -67,9 +69,15 @@ const cat = new Cat('Kitty', "female");
 const man = new Human('John', "male", "Hello Jenny!");
 const woman = new Human('Jenny', "female", "Hi John!");
 
-const catWoman = new CatWoman('Jenny', "female");
+const catWoman = new CatWoman('JennyCat', "female");
 
-const persons = [dog, cat, man, woman, catWoman];
+
+Cat.sayingPhrase('mi');
+
+const catWomanTwo = new CatWoman('Cat', "female");
+const catTwo = new Cat('Ben', 'male');
+
+const persons = [dog, cat, man, woman, catWoman, catWomanTwo, catTwo];
 
 function createString(persons) {
    return persons.map((person) => person.createString()).join('\n');
@@ -97,5 +105,4 @@ function createString(persons) {
 
 
 print(createString(persons));
-
 
--- a/submissions/Oleksandr-Heleta/OOP_exercise/index.js
+++ b/submissions/Oleksandr-Heleta/OOP_exercise/index.js
@@ -1,0 +1,69 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitants {
+   constructor(species, name, gender, say) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.say = say;
+   }
+   createInhabitantsString() {
+      return Object.values(this).join('; ');
+   }
+}
+
+class Human extends Inhabitants {
+   constructor(species, name, gender, say, legs, hands) {
+      super(species, name, gender, say);
+      this.legs = legs;
+      this.hands = hands;
+   }
+}
+
+class Animal extends Inhabitants {
+   constructor(species, name, gender, say, paws) {
+      super(species, name, gender, say);
+      this.paws = paws;
+
+   }
+}
+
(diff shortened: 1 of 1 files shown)


Oleksandr-Heleta:
I am still waiting review(

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -0,0 +1,85 @@
...
+   }
+}
+
+const dog = new Dog('Bob', "male", "bark");
+const cat = new Cat('Kitty', "female", "meow");
+
+const man = new Human('John', "male", "Hello Jenny!");
+const woman = new Human('Jenny', "female", "Hi John!");
+
+const catWoman = new Human('Jenny', "female", cat.say);
```
OleksiyRudenko, line 51: Nice try here. But this way catWoman uses saying of a particular cat. Besides it will be just copied.

What is needed is that all cat-women speak like any cat and when cat's standard speak gets changed, so do cat-women.

Static class properties/methods are to the rescue.

Oleksandr-Heleta:
Could You review one more!

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -31,32 +41,38 @@ class Animal extends Inhabitants {
...
+      super('cat', name, gender, "meow");
    }
 }
 
-const dog = new Dog('Bob', "male", "bark");
-const cat = new Cat('Kitty', "female", "meow");
+class CatWoman extends Human {
+   constructor(name, gender) {
+      super(name, gender)
+      this.mutation = new Cat();
```
OleksiyRudenko, line 58: This way you create a cat within a cat-woman. Doesn't reflect how it works in real world.

The following from the previous review is still valid.
> Static class properties/methods are to the rescue.

Please work on comments in their entirety and ask specific questions if anything is not clear.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -14,13 +14,23 @@ class Inhabitants {
       this.gender = gender;
       this.legs = legs;
       this.say = say;
+      this.propertiesOrder = ['species', 'name', 'gender', 'legs', 'say']
+   }
+
+   addProperty(property, order) {
+      this.propertiesOrder.splice(order, 0, property);
+   }
```
OleksiyRudenko, line 22: The following is still valid:

> The point of encapsulation principle of OOP is that every class handles its own properties.
> Let's implement class methods that produce class instance presentation string.

> You will employ method reloading and explicit inherited method version calling via super.

> The original project and task materials contain links to relevant sources that you will find helpful.

What is problem here? If a child class changes the way its property is implemented (e.g. we want to store now hands as an array of left and right hands) the code in a base class would break. This means we should change something in the base class to properly handle a property owned by a child class. This breaks the encapsulation principle. Let every class handle the properties it owns solely. JS provides required mechanisms. See hints in the comment above.

OleksiyRudenko (changes requested):
@Oleksandr-Heleta nice try.
It took the reviewer some time to provide hints that were ignored to a great extent. Implementation following the hints would make the code OOP/SOLID compliant.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -47,16 +45,20 @@ class Dog extends Animal {
...
 }
 
 class CatWoman extends Human {
    constructor(name, gender) {
-      super(name, gender)
-      this.mutation = new Cat();
-      this.say = this.mutation.say
+      super(name, gender, Cat.say)
    }
 }
```
Oleksandr-Heleta, line 64: Added static property and static method

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -14,23 +14,21 @@ class Inhabitants {
...
-   }
-
-   addProperty(property, order) {
-      this.propertiesOrder.splice(order, 0, property);
    }
 
    createString() {
-      return this.propertiesOrder.map((property) => this[property] || 'none');
+      return `${this.species}, ${this.name}, ${this.gender}, ${this.legs}, ${this.say};`;
    }
```
Oleksandr-Heleta, line 21: Changed render method

Oleksandr-Heleta (commented):
@OleksiyRudenko
I'm sorry for taken your time. I hope I understood you clear. Please, review one more!

submissions/Oleksandr-Heleta/OOP_exercise/index.js
OleksiyRudenko: This is the right direction. But cat-woman's saying is still copied from a primitive.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
OleksiyRudenko: Fine. Now, make this frgament DRYer using `Array#join` method.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+class Human extends Inhabitants {
+   constructor(name, gender, say) {
+      super('human', name, gender, 2, say);
+      this.hands = 2;
+   }
+
+   createString() {
+      return `${this.species}, ${this.name}, ${this.gender}, ${this.legs}, ${this.hands}, ${this.say};`;
+   }
```
OleksiyRudenko, line 32: You did method overloading but you do not benefit from explicit inherited method version call.
This wil make code dryer and hence less error prone.
Check learning materials.
Read comments in their entirety.
Ask questions if anything is not clear.

OleksiyRudenko (changes requested):
@Oleksandr-Heleta quite an improvement.
But still half-way.

Oleksandr-Heleta:
@OleksiyRudenko
I tried to fix. If cat-woman doesn't have linked say property, could you tell me how to do it

submissions/Oleksandr-Heleta/OOP_exercise/index.js
OleksiyRudenko: Now the method name promises to create a string, but it creates something else.
Yet there is way to keep the promise.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -28,7 +28,9 @@ class Human extends Inhabitants {
    }
 
    createString() {
-      return `${this.species}, ${this.name}, ${this.gender}, ${this.legs}, ${this.hands}, ${this.say};`;
+      const superArr = [...super.createString()];
+      superArr.splice(4, 0, this.hands);
```
OleksiyRudenko, line 32: I see what you mean and why changed the `createString` behaviour.
Might work. Although not really scalable.
What if base class developer (imagine you are on a big team here) changes the order of properties?
This will ruin your logic here.
So we either do not rely on base class implementation or we want the base class to be able to provide us with a info about where it is safe to insert extra items at.

submissions/Oleksandr-Heleta/OOP_exercise/index.js
```diff
@@ -47,18 +49,19 @@ class Dog extends Animal {
 class Cat extends Animal {
    static say = 'moew';
    constructor(name, gender) {
-      super('cat', name, gender, Cat.say);
-
+      super('cat', name, gender);
+      this.say = Cat.sayingPhrase();
```
OleksiyRudenko, line 53: `Cat.sayingPhrase` will return a primitive.
Look at getters/setters - they are quite interesting.... things.
I realize this task is untrivial. We can skip it. Understanding of untriviality and what doesn't really work in OOP (or at least doesn't have obvious solutions) is a great learning per se. So feel free keeping the code here as it is and dive deeper some time later.

Oleksandr-Heleta:
@OleksiyRudenko
I tried to use getter and setter, but I don't understand how it solve problem. It is like my first try, but with private property for cat.
Could you explain how solve this problem, Please!!!

Oleksandr-Heleta:
I hope I found the solution) 
Please, review

OleksiyRudenko (approved):
@Oleksandr-Heleta great job!

PR #416, opened 2022-09-03
Oop exercise

OlexiyDobroskok:
### OOP exercise

[Demo](https://olexiydobroskok.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/OlexiyDobroskok/a-tiny-JS-world/tree/OOP)

Please, review.

### DIFF
--- a/submissions/OlexiyDobroskok/OOP_exercise/index.js
+++ b/submissions/OlexiyDobroskok/OOP_exercise/index.js
@@ -26,12 +26,12 @@ class Inhabitant {
   }
 
   introduceYourSelf() {
-    const template = `<strong>${this.saying}! My name is ${this.name}. Species: ${this.species}. Gender: ${this.gender}.`;
+    const template = `<strong>${this.saying}! My name is ${this.name}. Species: ${this.species}. Gender: ${this.gender}.</strong>`;
     if (this.friends.length !== 0) {
       const friendsName = this.friends.map((friend) => friend.name).join(", ");
-      return template + `Friends: ${friendsName}.`;
+      return template + `<strong>Friends: ${friendsName}.</strong>`;
     }
-    return template + `Friends: Looking for friends!`;
+    return template + `<strong>Friends: Looking for friends!</strong>`;
   }
 }
 
--- a/submissions/OlexiyDobroskok/OOP_exercise/index.js
+++ b/submissions/OlexiyDobroskok/OOP_exercise/index.js
@@ -1,0 +1,85 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/OlexiyDobroskok/a-tiny-JS-world
+   Web app: https://olexiydobroskok.github.io/a-tiny-JS-world/
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+
+class Resident {
+  constructor(species, name, gender, saying, friends) {
+    this.species = species;
+    this.gender = gender;
+    this.friends = friends;
+    this.name = name;
+    this.saying = saying;
+  }
+  printPresentation() {
+    const properties = [
+      this.species,
+      this.name,
+      this.gender,
+      this.legs,
+      this.paws,
+      this.hands,
+      this.saying,
+      this.friends,
+    ];
+    const activeProperties = properties.filter(
+      (prop) => prop !== undefined && prop !== ""
+    );
+    print(activeProperties.join("; "));
+  }
+}
+
+class Human extends Resident {
+  constructor(name, gender, saying, friends) {
+    super("human", name, gender, saying, friends);
+    this.hands = 2;
+    this.legs = 2;
+  }
+}
+
+class Pet extends Resident {
+  constructor(species, name, gender, saying, friends) {
+    super(species, name, gender, saying, friends);
+    this.paws = 4;
+  }
+}
+
+class Cat extends Pet {
+  constructor(name, gender, saying, friends) {
+    super("cat", name, gender, saying, friends);
+  }
+}
+
+class Dog extends Pet {
+  constructor(name, gender, saying, friends) {
+    super("dog", name, gender, saying, friends);
+  }
+}
+
+class CatWoman extends Human {
+  constructor(name, gender, saying, friends) {
+    super(name, gender, saying, friends);
+  }
+}
+
+const man = new Human("José", "male", "Hola, amigo!", ["Nerea", "Lalo"]);
+const woman = new Human("Martina", "female", "Buenos días!", "Pako");
+const cat = new Cat("Lalo", "female", "¡miau miau!", ["José", "Nerea"]);
+const dog = new Dog("Pako", "male", "¡guau guau!", "Martina");
+
+const catwoman = new CatWoman("Nerea", "female", cat.saying + " muchachos!", [
+  "José",
+  "Lalo",
+]);
+
+function showATinyWorldResidents(residents) {
+  residents.forEach((resident) => resident.printPresentation());
+}
+
+const aTinyWorldResidents = [man, woman, cat, dog, catwoman];
+
+showATinyWorldResidents(aTinyWorldResidents);
--- a/submissions/OlexiyDobroskok/OOP_exercise/index.js
+++ b/submissions/OlexiyDobroskok/OOP_exercise/index.js
@@ -1,12 +1,3 @@
-/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
-   Complete the below for code reviewers' convenience:
-
-   Code repository: https://github.com/OlexiyDobroskok/a-tiny-JS-world
-   Web app: https://olexiydobroskok.github.io/a-tiny-JS-world/
-   */
-
-// ======== OBJECTS DEFINITIONS ========
-
 class Resident {
   constructor(species, name, gender, saying) {
     this.species = species;
--- a/submissions/OlexiyDobroskok/OOP_exercise/index.js
+++ b/submissions/OlexiyDobroskok/OOP_exercise/index.js
@@ -1,12 +1,3 @@
-/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
-   Complete the below for code reviewers' convenience:
-
-   Code repository: https://github.com/OlexiyDobroskok/a-tiny-JS-world
-   Web app: https://olexiydobroskok.github.io/a-tiny-JS-world/
-   */
-
-// ======== OBJECTS DEFINITIONS ========
-
 class Resident {
   constructor(species, name, gender, saying, friends) {
     this.species = species;
--- a/submissions/OlexiyDobroskok/OOP_exercise/index.js
+++ b/submissions/OlexiyDobroskok/OOP_exercise/index.js
@@ -1,8 +1,16 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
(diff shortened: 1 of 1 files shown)


OlexiyDobroskok:
Made a self-test and looked at the task from a new perspective)

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,159 @@
...
+    gender,
+    saying,
+    legs,
+    hands,
+    friends,
+  } = humanDefaultProperties) {
+    super({ species: "human", name, gender, saying, friends });
+    this.hands = hands ?? humanDefaultProperties.hands;
+    this.legs = legs ?? humanDefaultProperties.legs;
+  }
```
OleksiyRudenko, line 59: Why not just to set default value the same way it is done for `species`?
This is not a rhetoric question.

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,159 @@
...
+    this.name = name ?? inhabitantDefaultProperties.name;
+    this.gender = gender ?? inhabitantDefaultProperties.gender;
+    this.saying = saying ?? inhabitantDefaultProperties.saying;
+    this.friends = friends ?? inhabitantDefaultProperties.friends;
+    Inhabitant.worldPopulation.push(this);
+  }
+
+  addFriend({ name }) {
+    this.friends += name + "; ";
+  }
```
OleksiyRudenko, line 37: From the requirements:
> Friends list is a list of objects refs rather than names (strings)

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,159 @@
...
+      friends,
+    });
+    Dog.bark = this.saying;
+  }
+}
+
+class CatWoman extends Woman {
+  constructor({ name, saying = "", friends, legs, hands }) {
+    super({ name, saying, friends, legs, hands });
+    this.saying = Cat.meow + " " + saying;
```
OleksiyRudenko, line 132: Use static property is a good start in the right direction.
But here is the problem: if we change cat's defaut meow (which is allowed by Cat class API), cat-women's saying will be out of sync. Can you devise a way to bind their saying more tightly? You are only one step away from the solution.

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,159 @@
...
+};
+
+const humanDefaultProperties = {
+  hands: 2,
+  legs: 2,
+};
+
+const petsDefaultProperties = {
+  paws: 4,
+};
```
OleksiyRudenko, line 16: It would be a better approach to have defaults within class definitions. Not necessarily in form of explicit const definitions. KISS

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,159 @@
...
+    gender,
+    saying,
+    friends,
+  } = inhabitantDefaultProperties) {
+    this.species = species ?? inhabitantDefaultProperties.species;
+    this.name = name ?? inhabitantDefaultProperties.name;
+    this.gender = gender ?? inhabitantDefaultProperties.gender;
+    this.saying = saying ?? inhabitantDefaultProperties.saying;
+    this.friends = friends ?? inhabitantDefaultProperties.friends;
+    Inhabitant.worldPopulation.push(this);
```
OleksiyRudenko, line 32: Not a good idea to populate world from within inhabitants.
There is a Factory pattern in SW design, but this is not the case.

OleksiyRudenko (changes requested):
@OlexiyDobroskok a really good job!
A few improvements and we are done.

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: I’ve overdone the default values a little bit. =))

OlexiyDobroskok: Done

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: Done

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: done

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: Can I create a separate function that will create and push objects into an array?

OlexiyDobroskok: > Not a good idea to populate world from within inhabitants. There is a Factory pattern in SW design, but this is not the case.

Can I create a separate function that will create and push objects into an array?

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: @OleksiyRudenko The easiest way I see is to assign a default value to the static property of the cat:
Cat.meow = "¡miau miau";
    this.saying = Cat.meow;
but then all cats lose their identity and it makes me sad =)))

OleksiyRudenko: As long as property holds a primitive value the tie is weak, i.e. simply copied once. Strings, numbers and booleans are primitives. We need something that is passed by reference. These are JS objects (arrays and functions are also objects). Class methods can also be static. There are also very specific methods - getters and setters.
Given the above find a solution that will help to both meet requirements and minimize impact on code to keep it readable (i.e. we do not want dogs and humans to have added complexity just becase we need to resolve a matter between cats and cat-women).

submissions/OlexiyDobroskok/OOP_exercise/index.js
OleksiyRudenko: Yeah. Creating `World` class would be an overkill at this stage of project development.
But please avoid `.push`. `Array` methods provide more semantic ways to solve this.
And you will probably want to help inhabitants make friends after they are born. At least exisiting inhabitants cannot make friends with those who are not born yet.

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: @OleksiyRudenko Tried to bind Woman and cat by static method. Looks strange)

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: I abandoned this idea because it was originally intended to simplify the creation of the world, but Later there are problems with withdrawal of friends.What method can you recommend to replace the push?It seemed to me very convenient to work initially with an empty array. (

OleksiyRudenko: I reviewed this fragment. I agree that `.push` is a way to go.
The only thing remains: it is a world that consists of Inhabitants rather than world is a part of any Inhabitants.
Sure, syntax allows us to do this and make a static prop looks attractive. But having world as a part of inhabitant is counter-intuitive.
World can be a collection of inhabitants. Simplest collection is JS array. If we wanted some world-wide operations we would have multiple functions for those operations and it would make sense to scope them under class World.

submissions/OlexiyDobroskok/OOP_exercise/index.js
OleksiyRudenko: It does. Truth is that hybrids do not fit OOP hierarchical nature really well.
Composition is a better pattern in similar cases.
The purpose of this part of task was exactly to see the weirdness of how it has to be implemented in OOP paradigm.

OlexiyDobroskok: @OleksiyRudenko Good task. At a glance nothing difficult, but on the small details forces to revise a lot of materials). Do I still have to make some adjustments to the task?

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,138 @@
...
+    this.friends = friends;
+  }
+
+  addFriend(person) {
+    this.friends.push(person);
+  }
+
+  introduceYourSelf() {
+    if (this.friends.length === 0) {
+      this.friends = "Looking for friends!";
```
OleksiyRudenko, line 22: What will happen if we `.addFriend` after first call of `.introduceYourSelf`?

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,138 @@
...
+      return `<strong>${this.saying}! My name is ${this.name}. Species: ${this.species}. Gender: ${this.gender}. Friends: ${this.friends} </strong>`;
+    }
+    const friendsName = this.friends.map((friend) => friend.name).join(", ");
+    return `<strong>${this.saying}! My name is ${this.name}. Species: ${this.species}. Gender: ${this.gender}. Friends: ${friendsName}. </strong>`;
+  }
```
OleksiyRudenko, line 27: This fragment can be made DRY.
Optimize this method.

OleksiyRudenko (changes requested):
@OlexiyDobroskok nearly there.

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: Instead of an empty array, I get a string...)

OlexiyDobroskok: @

> What will happen if we `.addFriend` after first call of `.introduceYourSelf`?

@OleksiyRudenko Instead of an empty array, I get a string...) Done)

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: > This fragment can be made DRY. Optimize this method.

done

submissions/OlexiyDobroskok/OOP_exercise/index.js
```diff
@@ -0,0 +1,146 @@
...
+    this.friends.push(person);
+  }
+
+  introduceYourSelf() {
+    const template = `<strong>${this.saying}! My name is ${this.name}. Species: ${this.species}. Gender: ${this.gender}.</strong>`;
+    if (this.friends.length !== 0) {
+      const friendsName = this.friends.map((friend) => friend.name).join(", ");
+      return template + `<strong>Friends: ${friendsName}.</strong>`;
+    }
+    return template + `<strong>Friends: Looking for friends!</strong>`;
```
OleksiyRudenko, line 34: You have something for any situation about friends (0 or non-zero number of friends).
`const friendsName = this.friends.length ? this.friends.map((friend) => friend.name).join(", "); : "Looking for friends"`
Now you can have a single `return ` statement. Altogether 2 lines of code for this method.
Feel free searching for ternary operators and/or seek for explanation in Students' chat.

Ternaries are helpful, you will use them often. Worth mastering them.

OleksiyRudenko (approved):
@OlexiyDobroskok well done!

submissions/OlexiyDobroskok/OOP_exercise/index.js
OlexiyDobroskok: @OleksiyRudenko Thanks. Looks cool) Need to look at ternary operators.

PR #433, opened 2022-09-04
Add tiny-js-world task

misshatsker:
### Tiny JS World

[Demo](https://misshatsker.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/misshatsker/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/misshatsker/tiny-js-world/index.js
+++ b/submissions/misshatsker/tiny-js-world/index.js
@@ -1,0 +1,57 @@
+// ======== OBJECTS DEFINITIONS ========
+
+const human = {
+   species: 'human',
+   legs: 2,
+   hands: 2
+};
+
+const animal = {
+   species: 'animal',
+   legs: 4,
+   hands: 0
+};
+
+const woman = {
+   ...human,
+   name: 'Lilu',
+   gender: 'female',
+   saying: 'Aloha!'
+};
+
+const man = {
+   ...human,
+   name: 'Bob',
+   gender: 'male',
+   saying: 'Have a nice day!'
+};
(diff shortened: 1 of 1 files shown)


submissions/misshatsker/tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+const inhabitants = [
+   woman,
+   man,
+   dog,
+   cat
+];
+
+function getInhabitantInfo(inhabitant) {
+   return `${inhabitant.species}; ${inhabitant.name}; ${inhabitant.gender}; ${inhabitant.legs}; ${inhabitant.hands}; ${inhabitant.saying}`;
```
artem-trubin, line 51: This is a pattern here, the code is repeated. Imagine a situation where you would have more properties and you had a task to change the separator from `;` to, let's say, `,`. Find a way to specify keys by hand and then cycle through them getting information from objects. Make sure to **NOT** use `Object.values` and similar methods as they aren't reliable in terms of key order and presence.

artem-trubin (changes requested):
Hey, nice job. Though you went a little ahead and remade your demo for the OOP version already, I'm sure the pre-oop demo worked just fine. There is still one improvement there is to make here, check the comment below.

submissions/misshatsker/tiny-js-world/index.js
misshatsker: Thanks for the review! I've improved it, Could you please check it again?

artem-trubin (approved):
Great job!

PR #436, opened 2022-09-04
Oop classes

MarharytaBoichenko:
### OOP Exercise

[Demo](https://marharytaboichenko.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/MarharytaBoichenko/a-tiny-JS-world/blob/master/index.js)

Please, review.


MarharytaBoichenko:
Sorry, need help, I can`t  understand why  I have such  problem  with  this pull request  and commits(( 
Where  was the first  mistake why I have  all  this files in my branch

OleksiyRudenko:
Check the bot's suggestions and items 2 and 3 [here](https://github.com/kottans/frontend-2022-homeworks#%D0%B4%D0%BE%D0%B2%D1%96%D0%B4%D0%BA%D0%BE%D0%B2%D1%96-%D0%BC%D0%B0%D1%82%D0%B5%D1%80%D1%96%D0%B0%D0%BB%D0%B8-%D1%82%D0%B0-%D1%83%D1%81%D1%83%D0%BD%D0%B5%D0%BD%D0%BD%D1%8F-%D0%BF%D1%80%D0%BE%D0%B1%D0%BB%D0%B5%D0%BC)

MarharytaBoichenko:
Thanks for great instruction! My problem was in  committing to  main branch  directly,  but it seems I did  all this  steps in the  instruction  to  the wrong  commit( 
I took commit 8d092ecc23a4dbaa3dcf9c800c49610892c816d5,  the  last in Popup, because  in Dom task I committed to  main   but now  I see  all  popup commits  here( 
It seems I  can-t understand what commit I  need  for this  step -  " identify the most recent commit that both usptream and origin share (c4ebbab in our example)"

OleksiyRudenko:
>  It seems I can-t understand what commit I need for this step - " identify the most recent commit that both usptream and origin share (c4ebbab in our example)"

You just look for a commit that is present on both upstream and fork histories. To see the upstream's history you'd need to clone it separately.

MarharytaBoichenko:
I checked my  code according to the recomendations, please  review

submissions/MarharytaBoichenko/oop/index.js
```diff
@@ -0,0 +1,78 @@
...
+    this.name = name;
+    this.species = species;
+    this.friends = friends;
+    this.saying = saying;
+    this.properties = ["name", "species", "friends", "saying"];
+  }
+  toPrint() {
+    return this.properties
+      .map((prop) => {
+        const neededProp = this[prop] === undefined ? `no ${prop}` : this[prop];
```
OleksiyRudenko, line 12: Use `Array#filter` to keep props that make sense, then stringify what remains with `Array#map`.

submissions/MarharytaBoichenko/oop/index.js
```diff
@@ -0,0 +1,78 @@
...
+class Woman extends Person {
+  constructor(name, friends, saying) {
+    super(name, "woman", friends, saying);
+  }
+}
+class Animal extends Inhabitant {
+  constructor(name, species, legs, gender, friends, saying) {
+    super(name, species, friends, saying);
+    this.legs = legs;
+    this.gender = gender;
```
OleksiyRudenko, line 44: What makes `legs` and `gender` different from `name` in terms of handling and generalization of shared properties?

submissions/MarharytaBoichenko/oop/index.js
```diff
@@ -0,0 +1,78 @@
...
+    this.properties = [...this.properties, "gender", "legs"];
```
OleksiyRudenko, line 45: A child class shouldn't directly change inherited property.
A base class shouldn't handle child's properties.
Altogether, every class should handle end-to-end the properties it owns immediately.
Use method overloading and explicit call to inherited method version.

OleksiyRudenko (changes requested):
@MarharytaBoichenko good job!
I wonder why having suboptimal implementations that break OOP/SOLID principles?

submissions/MarharytaBoichenko/oop/index.js
MarharytaBoichenko: ok

submissions/MarharytaBoichenko/oop/index.js
MarharytaBoichenko: I changed  gender, but not  legs,  I think  it differs  because here class Animal extends Inhabitant   in parents  class Inhabitant  there  no  legs.

OleksiyRudenko: Why `legs` do not deserve the same as `name`?

MarharytaBoichenko: I think  my  mistake is that  I write legs for  Animal  but thought about legs  as  differnt properties  ,  not as legs  which  Person has  ). As  I  understand  if  it legs  and this  property  is  in Animal  and in  Person  too so we  need handle it  as  name and  gender  and legs  will go  from  parent  class.  And  if I decide  that legs  in class  Animals  is  diffferent property  - paws so in Animal it will be this.paw =  paw . Is  it  right?

OleksiyRudenko: Exactly

submissions/MarharytaBoichenko/oop/index.js
OleksiyRudenko: Still an issue.
Use method overloading/overriding and `super` API to call inherited method version. `super` has other useful applications, not only to call parent's class constructor.

submissions/MarharytaBoichenko/oop/index.js
MarharytaBoichenko: thanks, done

submissions/MarharytaBoichenko/oop/index.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+class Woman extends Person {
+  constructor(name, friends, saying) {
+    super(name, "woman", friends, saying);
+  }
+
+  toPrint() {
+    return super.toPrint();
+  }
```
OleksiyRudenko, line 49: No need to overload a method to only explicitly call an inherited version of the method.

submissions/MarharytaBoichenko/oop/index.js
```diff
@@ -0,0 +1,102 @@
...
+class Catwoman extends Cat {
+  constructor(name, paws, gender, hands, friends) {
+    super(name, "catwoman", paws, gender, friends);
+    this.hands = hands;
+  }
+
+  toPrint() {
+    return super.toPrint() + `, my hands: ${this.hands}`;
+  }
+}
```
OleksiyRudenko, line 90: Notice, how hybrids break OOP principles: it forces you to duplicate properties.
Although, there are other approaches to achieve this without code duplication, the major learning here should be that when you need to compose properties and behaviours in a non-strictly-hiearchical manner, OOP won't work, and composition pattern is a way to go. Keep in mind and when you face a need to build hybrids, read about and use composition.

OleksiyRudenko (approved):
@MarharytaBoichenko great job!

PR #439, opened 2022-09-05
tiny-js-oop

Yasya23:
### OOP classes

[Demo](https://yasya23.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Yasya23/a-tiny-JS-world/tree/gh-pages)

Please, review.

### DIFF
--- a/submissions/Yasya23/tiny-js-oop/index.js
+++ b/submissions/Yasya23/tiny-js-oop/index.js
@@ -1,0 +1,61 @@
+class Inhabitants {
+  constructor(species, name, gender) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+  }
+
+  showProperties() {
+    return `${this.species}; ${this.name}; ${this.gender}`;
+  }
+}
+
+class Human extends Inhabitants {
+  constructor(name, gender, saying) {
+    super("human", name, gender);
+    this.hands = 2;
+    this.legs = 2;
+    this.saying = saying;
+  }
+
+  showProperties() {
+    return `${super.showProperties()}; ${this.hands}; ${this.legs}; ${
+      this.saying
+    }`;
+  }
+}
+
+class Animal extends Inhabitants {
+  constructor(species, name, gender, saying) {
+    super(species, name, gender);
+    this.paws = 4;
+    this.saying = saying;
+  }
+  showProperties() {
+    return `${super.showProperties()}; ${this.paws}; ${this.saying}`;
+  }
+}
+
+class Cat extends Animal {
+  constructor(name, gender, saying) {
+    super("cat", name, gender, saying);
+  }
+}
+
+class Dog extends Animal {
+  constructor(name, gender, saying) {
+    super("dog", name, gender, saying);
+  }
+}
+
+const dog = new Dog("Bob", "male", "woof");
+const cat = new Cat("Grey", "male", "meow");
+const catWoman = new Cat("Lily", "female", "meow");
+const man = new Human("Anton", "male", "Have a good day!");
+const woman = new Human("Anna", "female", "Good luck!");
+
+const inhabitants = [dog, cat, catWoman, man, woman];
+
+inhabitants.forEach(function (inhabitant) {
+  print(inhabitant.showProperties());
+});
--- a/submissions/Yasya23/tiny-js-oop/index.js
+++ b/submissions/Yasya23/tiny-js-oop/index.js
@@ -1,38 +1,35 @@
 class Inhabitants {
-  constructor(species, name, gender) {
+  constructor(species, name, gender, saying) {
     this.species = species;
     this.name = name;
     this.gender = gender;
+    this.saying = saying;
   }
 
   showProperties() {
-    return `${this.species}; ${this.name}; ${this.gender}`;
+    return [this.species, this.name, this.gender];
   }
 }
 
 class Human extends Inhabitants {
   constructor(name, gender, saying) {
-    super("human", name, gender);
+    super("human", name, gender, saying);
     this.hands = 2;
     this.legs = 2;
-    this.saying = saying;
   }
 
   showProperties() {
-    return `${super.showProperties()}; ${this.hands}; ${this.legs}; ${
-      this.saying
-    }`;
+    return [...super.showProperties(), this.hands, this.legs, this.saying];
   }
 }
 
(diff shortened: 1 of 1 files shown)


Yasya23:
Self-check has been done. Everything should be ok.

Yasya23:
Self-check has been done.

submissions/Yasya23/tiny-js-oop/index.js
```diff
@@ -0,0 +1,61 @@
...
+      this.saying
+    }`;
+  }
+}
+
+class Animal extends Inhabitants {
+  constructor(species, name, gender, saying) {
+    super(species, name, gender);
+    this.paws = 4;
+    this.saying = saying;
```
OleksiyRudenko, line 32: What makes `saying` different from `name` in terms of generalization into the base class?

submissions/Yasya23/tiny-js-oop/index.js
```diff
@@ -0,0 +1,61 @@
+class Inhabitants {
+  constructor(species, name, gender) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+  }
+
+  showProperties() {
+    return `${this.species}; ${this.name}; ${this.gender}`;
```
OleksiyRudenko, line 9: This and similar fragments below:
Benefit from `Array` methods. This will make code DRYer.

OleksiyRudenko (changes requested):
@Yasya23 good job!
A couple of improvements and we are done.

submissions/Yasya23/tiny-js-oop/index.js
Yasya23: Hello, thanks for checking! There is no difference in these cases, fixed it.

submissions/Yasya23/tiny-js-oop/index.js
Yasya23: Corrected for an array.

submissions/Yasya23/tiny-js-oop/index.js
```diff
@@ -57,5 +54,5 @@ const woman = new Human("Anna", "female", "Good luck!");
 const inhabitants = [dog, cat, catWoman, man, woman];
 
 inhabitants.forEach(function (inhabitant) {
-  print(inhabitant.showProperties());
+  print(inhabitant.showProperties().join("; "));
 });
```
OleksiyRudenko, line 58: Arrow functions will make code less verbose.
Compare:
```
inhabitants.forEach(function (inhabitant) {
  print(inhabitant.showProperties().join("; "));
});

inhabitants.forEach(inhabitant => print(inhabitant.showProperties().join("; ")));
```

submissions/Yasya23/tiny-js-oop/index.js
```diff
@@ -1,38 +1,35 @@
...
     this.hands = 2;
     this.legs = 2;
-    this.saying = saying;
   }
 
   showProperties() {
-    return `${super.showProperties()}; ${this.hands}; ${this.legs}; ${
-      this.saying
-    }`;
+    return [...super.showProperties(), this.hands, this.legs, this.saying];
```
OleksiyRudenko, line 22: Here and line 32 below:
- repetition
- also a shared property should be handled by the class that owns it

OleksiyRudenko (changes requested):
@Yasya23 nearly there

submissions/Yasya23/tiny-js-oop/index.js
Yasya23: Thanks for checking. Yes, it looks much better, fixed it.

submissions/Yasya23/tiny-js-oop/index.js
Yasya23: Fixed it but I`m not sure about my decision. If it's incorrect please give me advice.

OleksiyRudenko: That's a good start.
Your previous method version returned a string. Now it returns array not joined into string. Why?

Yasya23: I'm a little confused. The join method returns a string from the array of elements. If I add it here then it will return a string.

`  showProperties() {
    return [
      ...super.showProperties().slice(0, 3),
      this.paws,
      ...super.showProperties().slice(3, 4),
    ].join("; ");
  }`

But I don’t understand the difference, because at the end I still iterate over the array and add a join. 

`inhabitants.forEach(inhabitant => print(inhabitant.showProperties().join("; ")));`

Apparently, there is something I don't understand.

OleksiyRudenko: My question was about reasons behind changing from string to array. It is a significant change.

OleksiyRudenko: Both approaches may be valid. Want to understand reasoning.

Yasya23: Because of that post)

This and similar fragments below:
Benefit from Array methods. This will make code DRYer.


For myself, I realized that I need to create an array instead of a string. But now I understand that I could iterate through the array of values ​​​​using the map method and add a join, and not write a semicolon after each value. Something like this.

`["species", "name", "gender", "saying"]
  .map((property) => this.property)
  .join("; ");
`

Only working with an array for me here looks easier if I want to insert a value in its middle.

submissions/Yasya23/tiny-js-oop/index.js
```diff
@@ -0,0 +1,67 @@
...
+    this.hands = 2;
+    this.legs = 2;
+  }
+
+  showProperties() {
+    return [
+      ...super.showProperties().slice(0, 3),
+      this.hands,
+      this.legs,
+      ...super.showProperties().slice(3, 4),
```
OleksiyRudenko, line 26: I realize why slicing.
Just notice that this is not a really good approach. Why? A: This means that a developer working on `Human` or any other subclass would need to know what is the right order of props in the base class. Likewise, developer working on `Inhabitants` class (e.g. add a property between already existing) should be aware their changes may affect subclasses' behavior. On big projects it is not feasible. Ideally, every class handles its own properties and unaware of other classes internals.

P.S. BTW, class names by convention are singular unless they really contain multiple instances of anything.

OleksiyRudenko (approved):
@Yasya23 great job done!

PR #450, opened 2022-09-05
JS OOP Frogger Game Task

franchukv:
### JS OOP Frogger Game

[Demo](https://franchukv.github.io/js-oop-frogger/) | [Code base](https://github.com/franchukv/js-oop-frogger)

Fixed all bugs or defects which other students received:
1. Now enemies don't 'blink' when creating.
2. Now user visually can see that he crossed the road successfully.

Please, review.

### DIFF
--- a/submissions/franchukv/js-oop-frogger/app.js
+++ b/submissions/franchukv/js-oop-frogger/app.js
@@ -2,35 +2,18 @@ const CANVAS_HEIGHT = 606;
 const CANVAS_WIDTH = 505;
 const TILE_HEIGHT = 83;
 const TILE_WIDTH = 101;
-const ENEMY_SPEED = 80;
-const PLAYER_STAT = {
-    x: 202,
-    y: 380,
-    height: 40,
-    width: 40,
-};
 
 const Player = function () {
     this.sprite = 'images/char-boy.png';
-    this.x = PLAYER_STAT.x;
-    this.y = PLAYER_STAT.y;
-    this.height = PLAYER_STAT.height;
-    this.width = PLAYER_STAT.width;
-};
-
-Player.prototype.update = function () {
-    if (this.y < 0) {
-        this.resetPlayerPosition();
-    }
-};
-
-Player.prototype.render = function () {
-    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+    this.x = 202;
+    this.y = 380;
+    this.height = 40;
+    this.width = 40;
 };
 
 Player.prototype.resetPlayerPosition = function () {
-    this.x = PLAYER_STAT.x;
-    this.y = PLAYER_STAT.y;
+    this.x = 202;
+    this.y = 380;
 };
 
 Player.prototype.handleInput = function (key) {
@@ -52,16 +35,37 @@ Player.prototype.handleInput = function (key) {
     }
 };
 
-const Enemy = function (x, y, player) {
+Player.prototype.update = function () {
+    if (this.y < 0) {
+        this.resetPlayerPosition();
+    }
+};
+
+Player.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const Enemy = function (x, y, speed, player) {
     this.sprite = 'images/enemy-bug.png';
     this.x = x;
     this.y = y;
     this.width = 80;
     this.height = 40;
-    this.speed = ENEMY_SPEED;
+    this.speed = speed;
     this.player = player;
 };
 
+Enemy.prototype.checkCollision = function () {
+    if (
+        this.player.y > this.y - this.height &&
+        this.player.y - this.player.height < this.y &&
+        this.player.x + this.player.width > this.x &&
+        this.player.x < this.x + this.width
+    ) {
+
+        this.player.resetPlayerPosition();
+    }
+};
 
 Enemy.prototype.update = function (dt) {
     this.checkCollision();
@@ -77,33 +81,12 @@ Enemy.prototype.render = function () {
     ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 };
 
-Enemy.prototype.setStartEnemiesPosition = function () {
-    for (let i = 0; i < 3; i++) {
-        let y = i * TILE_HEIGHT + 50;
-        let x = i * -TILE_WIDTH;
-        allEnemies.push(new Enemy(x, y, player));
-    }
-};
-
-Enemy.prototype.checkCollision = function () {
-    if (
-        (
-            this.player.y > this.y - this.height &&
-            this.x + this.width > this.player.x &&
-            this.player.x + this.player.width > this.x &&
-            this.player.y - this.player.height < this.y
-        )
-    ) {
-
-        player.resetPlayerPosition();
-    }
-};
-
-const allEnemies = [];
-const enemy = new Enemy();
 const player = new Player();
-
-Enemy.prototype.setStartEnemiesPosition();
+const allEnemies = [
+    new Enemy(0, 55, 280, player),
+    new Enemy(0, 135, 150, player),
+    new Enemy(0, 220, 200, player)
+];
 
 document.addEventListener('keyup', function (e) {
     var allowedKeys = {
--- a/submissions/franchukv/js-oop-frogger/app.js
+++ b/submissions/franchukv/js-oop-frogger/app.js
@@ -52,13 +52,14 @@ Player.prototype.handleInput = function (key) {
     }
 };
 
-const Enemy = function (x, y) {
+const Enemy = function (x, y, player) {
     this.sprite = 'images/enemy-bug.png';
     this.x = x;
     this.y = y;
     this.width = 80;
     this.height = 40;
     this.speed = ENEMY_SPEED;
+    this.player = player;
 };
 
 
@@ -80,17 +81,17 @@ Enemy.prototype.setStartEnemiesPosition = function () {
     for (let i = 0; i < 3; i++) {
         let y = i * TILE_HEIGHT + 50;
         let x = i * -TILE_WIDTH;
-        allEnemies.push(new Enemy(x, y));
+        allEnemies.push(new Enemy(x, y, player));
     }
 };
 
 Enemy.prototype.checkCollision = function () {
     if (
         (
-            player.y > this.y - this.height &&
-            this.x + this.width > player.x &&
-            player.x + player.width > this.x &&
-            player.y - player.height < this.y
+            this.player.y > this.y - this.height &&
+            this.x + this.width > this.player.x &&
+            this.player.x + this.player.width > this.x &&
+            this.player.y - this.player.height < this.y
         )
     ) {
 
--- a/submissions/franchukv/js-oop-frogger/app.js
+++ b/submissions/franchukv/js-oop-frogger/app.js
@@ -14,12 +14,12 @@ const PLAYER_CONF = {
     }
 }
 
-const EMENIES_CONF = {
+const ENEMIES_CONF = {
     size: {
         height: 40,
         width: 80
     },
-    enemy: [{
+    enemies: [{
         x: 0,
         y: 55,
         speed: 280
@@ -78,13 +78,13 @@ Player.prototype.render = function () {
     ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 };
 
-const Enemy = function ({ size }, enemy, player) {
+const Enemy = function ({ size }, enemyProperties, player) {
     this.sprite = 'images/enemy-bug.png';
     this.height = size.height;
     this.width = size.width;
-    this.x = enemy.x;
-    this.y = enemy.y;
-    this.speed = enemy.speed;
+    this.x = enemyProperties.x;
+    this.y = enemyProperties.y;
+    this.speed = enemyProperties.speed;
     this.player = player;
 };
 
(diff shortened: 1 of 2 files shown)


franchukv:
Self-check done.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,116 @@
...
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+Enemy.prototype.setStartEnemiesPosition = function () {
+    for (let i = 0; i < 3; i++) {
+        let y = i * TILE_HEIGHT + 50;
+        let x = i * -TILE_WIDTH;
+        allEnemies.push(new Enemy(x, y));
+    }
+};
```
OleksiyRudenko, line 85: Class should handle its instances. Class is designated to handle a single instance.
Also from the check-list:
>  classes do not refer to any global variables, like global variable player, which is an instance of Player class
> (referring to global constants and globals provided by the gaming platform like Resources is OK);
> Hint: pass Player instance as an argument to every enemy.

The code has this issue in the fragments below.

OleksiyRudenko (changes requested):
@franchukv pleas double-check yourself against requirements.
Feel free asking specific questions in Students' chat if anything is not clear.

submissions/franchukv/js-oop-frogger/app.js
franchukv: With hope that I correctly understand this requirement, I fixed this. Please review again.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,117 @@
...
+
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+Enemy.prototype.setStartEnemiesPosition = function () {
+    for (let i = 0; i < 3; i++) {
+        let y = i * TILE_HEIGHT + 50;
+        let x = i * -TILE_WIDTH;
+        allEnemies.push(new Enemy(x, y, player));
```
OleksiyRudenko, line 84: The problem with `allEnemies` is exactly the same as it was with `player`.
You need to recognize the problems of the same class.
Not illegal but a bit wierd to let an enemy to add itself to a naturally external container.
We can call this behavior "enlist yourself in squad" here, but normally we want entities being decoupled from containers for them. This way we separate concerns (management of an individual vs management of a set of individuals).
You may keep this as an `Enemy`'s method, but need to solve global variable reference problem.

Also `player` here is a global variable. Do not expect mentors to red-pen problems of the same class.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,117 @@
...
+    if (
+        (
+            this.player.y > this.y - this.height &&
+            this.x + this.width > this.player.x &&
+            this.player.x + this.player.width > this.x &&
+            this.player.y - this.player.height < this.y
+        )
+    ) {
+
+        player.resetPlayerPosition();
```
OleksiyRudenko, line 98: Check your code for "illegal" global variables.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+};
+
+const allEnemies = [];
+const enemy = new Enemy();
```
OleksiyRudenko, line 103: WHy do we need this?

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,117 @@
+const CANVAS_HEIGHT = 606;
+const CANVAS_WIDTH = 505;
+const TILE_HEIGHT = 83;
+const TILE_WIDTH = 101;
+const ENEMY_SPEED = 80;
+const PLAYER_STAT = {
```
OleksiyRudenko, line 6: What does "STAT" mean?
Do not use abbreviations that are not conventional in software development.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,117 @@
...
+            break;
+    }
+};
+
+const Enemy = function (x, y, player) {
+    this.sprite = 'images/enemy-bug.png';
+    this.x = x;
+    this.y = y;
+    this.width = 80;
+    this.height = 40;
```
OleksiyRudenko, line 60: Looks like enemy and player share some features.

OleksiyRudenko (changes requested):
@franchukv need another iteration of fixes.

submissions/franchukv/js-oop-frogger/app.js
franchukv: Okay, now I understand what does mean 'illegal' variables in this context.  Rewrote app, please review again)

submissions/franchukv/js-oop-frogger/app.js
franchukv: It means "characteristics", just a wrong naming, fixed

submissions/franchukv/js-oop-frogger/app.js
franchukv: We don't need this, fixed

submissions/franchukv/js-oop-frogger/app.js
franchukv: Fixed

submissions/franchukv/js-oop-frogger/app.js
OleksiyRudenko: Setting constants is yet another way to keep similar things in sync.

P.S. Shared properties and behaviour is a good enough reason to have a base class to encapsulate behaviour and properties handling. We just do not focus on this in this task. Just "a really-really nice to have". In OOP Exercise there are no excuses.

franchukv: Okay, got it, thanks.

franchukv: Please review again, I rewrote the app using constants.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,134 @@
...
+    if (this.y < 0) {
+        this.resetPlayerPosition(PLAYER_CONF);
+    }
+};
+
+Player.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const Enemy = function ({ size }, enemy, player) {
```
OleksiyRudenko, line 81: Function declaration says that the function expects and enemy as one of the parameters.
Is it actually an instance of class `Enemy` expected here?

OleksiyRudenko (changes requested):
@franchukv pretty well done!
Let's add some clarity.
Code is written for other people.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,134 @@
...
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const player = new Player(PLAYER_CONF);
+const allEnemies = [
+    new Enemy(EMENIES_CONF, EMENIES_CONF.enemy[0], player),
+    new Enemy(EMENIES_CONF, EMENIES_CONF.enemy[1], player),
+    new Enemy(EMENIES_CONF, EMENIES_CONF.enemy[2], player),
+];
```
OleksiyRudenko, line 122: As a peer developer I am tasked to add 20 more enemies.
Adding 20 more nearly identical rows of code is a no-go.
Can we make this code DRYer?

OleksiyRudenko (changes requested):
@franchukv yet another thing.

submissions/franchukv/js-oop-frogger/app.js
franchukv: Yes, we can use loop into 'addEnemiesToInitArray' function to add instance of enemies from 'ENEMIES CONF.enemy' to 'allEnemies'.

OleksiyRudenko: Let's make it.

submissions/franchukv/js-oop-frogger/app.js
franchukv: If I understood the question right, the answer will be: actually no, we expect basis arguments of 'enemy' for creating an instance of class Enemy.

submissions/franchukv/js-oop-frogger/app.js
franchukv: Done!

submissions/franchukv/js-oop-frogger/app.js
franchukv: Anyway, renamed.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+function addEnemiesToInitArray() {
+    ENEMIES_CONF.enemies.forEach((enemy) => {
+        allEnemies.push(new Enemy(ENEMIES_CONF, enemy, player));
+    });
+}
```
OleksiyRudenko, line 121: Extension to the requirement re global variables. It is good to avoid global variables as much as possible.
Well designed function relies on explicitly passed parameters only.
Read about pure functions to learn why.
You do not need to fix this in this code.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+const TILE_HEIGHT = 83;
+const TILE_WIDTH = 101;
+
+const PLAYER_CONF = {
+    size: {
+        height: 40,
+        width: 40
+    },
+    position: {
+        x: 202,
```
OleksiyRudenko, line 12: You could calculate this from TILE_WIDTH programmatically. So whenever tile width changes, this param changes accordingly and automagically.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+};
+
+const ENEMIES_CONF = {
+    size: {
+        height: 40,
+        width: 80
+    },
+    enemies: [{
+        x: -80,
```
OleksiyRudenko, line 23: You could pass repeating value at `new Enemy` moment.
E.g. `new Enemy({ x: -80, ...ENEMY_CONF}, ....)`
Would make enemies config 33% shorter. This would also help to change the number in a single place should we need to do this.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+
+        this.player.resetPlayerPosition(PLAYER_CONF);
+    }
+};
+
+Enemy.prototype.update = function (dt) {
+    this.checkCollision();
+
+    if (this.x >= CANVAS_WIDTH) {
+        this.x = -80;
```
OleksiyRudenko, line 107: Constants help make code DRY and numbers semantically comprehensible.

OleksiyRudenko (changes requested):
@franchukv let's polish it in order to practise best practices in software development.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+    this.x += this.speed * dt;
+};
+
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+function addEnemiesToInitArray() {
+    ENEMIES_CONF.enemies.forEach((enemy) => {
+        allEnemies.push(new Enemy(ENEMIES_CONF, enemy, player));
```
OleksiyRudenko, line 119: btw, why do we send all enemies configuration to every enemy?
I see that `size` is extracted at receiving party. But is would much more obvious if we use the opportunity to send exactly what is needed

OleksiyRudenko: And `enemy` here, actually, is not an instance of `Enemy` but rather enemy physical parameters.
It would be great if this was clear from reading the code without need to scroll up to class definition to nderstand what is going on in this line of code. On big codebases it would be a real pain.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+    position: {
+        x: 202,
+        y: 380
+    }
+};
+
+const ENEMIES_CONF = {
+    size: {
+        height: 40,
+        width: 80
```
OleksiyRudenko, line 20: Any correlation with `-80`?

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
...
+        speed: 280
+    },
+    {
+        x: -80,
+        y: 135,
+        speed: 150
+    },
+    {
+        x: -80,
+        y: 220,
```
OleksiyRudenko, line 34: Can `y` be calculated from row number?

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,136 @@
+const CANVAS_HEIGHT = 606;
+const CANVAS_WIDTH = 505;
```
OleksiyRudenko, line 2: Looks like there is correlation between tile width and canvas dimensions. Let computer do the calcs.

submissions/franchukv/js-oop-frogger/app.js
```diff
@@ -0,0 +1,145 @@
...
+    ENEMIES_CONF.enemies.forEach((enemyProperty) => {
+        allEnemies.push(
+            new Enemy(ENEMIES_CONF.size, -tileWidth, enemyProperty, player)
+        );
+    });
+};
+
+const player = new Player(PLAYER_CONF);
+const allEnemies = [];
+createEnemies(ENEMIES_CONF, TILE_HEIGHT, TILE_WIDTH);
```
OleksiyRudenko, line 134: `const allEnemies = createEnemies(...)` would be more obvious as it would self-explain the effect of `createEnemies`

OleksiyRudenko (approved):
@franchukv great job!
Employ passing whatever a function needs to deal with as arguments to functions and return from functions whatever they do rather than directly use/change global variables.
It makes code cleaner.

PR #451, opened 2022-09-05
completed first approach

sofiichuk:
### a tiny js world: first approach

[Demo](https://sofiichuk.github.io/a-tiny-JS-world) |
[Code base](https://github.com/sofiichuk/a-tiny-JS-world/blob/populate-world/index.js)

Please, review.


sofiichuk:
checked

sofiichuk:
hi.
it does work indeed)
i checked.
although i could've change it by the means of inheritance (as i did here - https://github.com/sofiichuk/a-tiny-JS-world/blob/OOP-exercise/index.js).
but the goal was - to do it without learning OOP approach ( i was really unaware of) - and so did i.
so do i have to change anything yet?

artem-trubin:
Well yes, your catwoman does say what the cat *initially* says, but that's why I marked *ANYTIME* in my comment. Consider this:

```js
const cat = { saying: "First test" };
const catwoman = { saying: cat.saying };
console.log(catwoman.saying) // this will be "First test"
cat.saying = "Second test";
console.log(catwoman.saying) // this should be "Second test", but with your implementation, it will still be "First test"
```

So you need to find a way to *connect* those values in some way. And yes, for the means of this task, without using classes syntax.

sofiichuk:
spend so much time...
why hadn't you told me how to solve it?
fixed that finally

artem-trubin (approved):
GJ, it works, but please check my comment below and keep it in mind in the future.

submissions/sofiichuk/a-tiny-js-world/index.js
```diff
@@ -0,0 +1,57 @@
...
+const catWoman = {
+    species: 'chimera',
+    name: 'undefined',
+    gender: 'undefined',
+    legs: 'undefined',
+    hands: 'undefined',
+    friends: 'undefined'
+};
+
+Object.setPrototypeOf(catWoman, cat);
```
artem-trubin, line 50: This is the second time I see `Object.setPrototypeOf`, did anyone tell you to use it? [Specs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) say it is a very slow method and there is an alternative - `Object.create()`. Or you could've used `bind`. Anyway, keep these in mind.

sofiichuk: at first i tried to transform cat.saying to an object/array to pass it by reference;
then i tried to combine cat and catwoman props using object.assign;
i failed with both;
then i found out people handle this with a __proto__;
but i had already known that using __proto__ in an explicit way is strongly not recommended;
so i continued searchin for smth similar as a substitution;
nth-google-request was "set prototype" and it led to setPrototype method;
(and now i see it's no better than using __proto__ - but already afterwards)

artem-trubin: Got it, nice research)

PR #464, opened 2022-09-07
update tiny js world with classes

Yuliiadd:
### OOP Exercise

[Demo](https://yuliiadd.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Yuliiadd/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/YuliiaDyka/oop-exercise/index.js
+++ b/submissions/YuliiaDyka/oop-exercise/index.js
@@ -1,40 +1,44 @@
 class LivingBeing {
-   constructor(species, name, gender, saying) {
+   constructor(species, name, gender, saying, legs) {
       this.species = species;
       this.name = name;
       this.gender = gender;
       this.saying = saying;
+      this.legs = legs;
    };
 
    showProperties() {
-      return Object.getOwnPropertyNames(this).map(prop => this[prop]).join('; ');
+      return `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.legs}`;
    }
 };
 
 class Human extends LivingBeing {
-   constructor(name, gender, saying) {
-      super("Human", name, gender, saying);
-      this.legs = 2;
+   constructor(name, gender, saying, legs) {
+      super("Human", name, gender, saying, legs);
       this.hands = 2;
    };
+
+   showProperties() {
+      return super.showProperties() + `; ${this.hands}`;
+   }
+
 };
 
 class Animal extends LivingBeing {
-   constructor(species, name, gender, saying) {
-      super(species, name, gender, saying);
-      this.legs = 4;
+   constructor(species, name, gender, saying, legs) {
+      super(species, name, gender, saying, legs);
    };
 };
 
 class Cat extends Animal {
-   constructor(name, gender) {
-      super("Kitty", name, gender, "Meooow!");
+   constructor(name, gender, legs) {
+      super("Kitty", name, gender, "Meooow!", legs);
    };
 };
 
 class Dog extends Animal {
-   constructor(name, gender) {
-      super("Puppy", name, gender, "Woof!");
+   constructor(name, gender, legs) {
+      super("Puppy", name, gender, "Woof!", legs);
    };
 };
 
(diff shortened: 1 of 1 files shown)


Yuliiadd:
@OleksiyRudenko
Hi! I did self-checks. Please, review.

submissions/YuliiaDyka/oop-exercise/index.js
```diff
@@ -0,0 +1,60 @@
...
+   constructor(species, name, gender, saying) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.saying = saying;
+   };
+
+   showProperties() {
+      return Object.getOwnPropertyNames(this).map(prop => this[prop]).join('; ');
+   }
```
OleksiyRudenko, line 11: Requirements re `Object` methods are still valid.

Also read about (and use) methods overloading/overriding and extended `super` API (calling inherited methods versions). You may want to revisit JS OOP educational materials.

submissions/YuliiaDyka/oop-exercise/index.js
```diff
@@ -0,0 +1,60 @@
...
+      super("Human", name, gender, saying);
+      this.legs = 2;
+      this.hands = 2;
+   };
+};
+
+class Animal extends LivingBeing {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender, saying);
+      this.legs = 4;
```
OleksiyRudenko, line 25: Why `name` is defined in the base class and `legs` in all classes that have legs?

OleksiyRudenko (changes requested):
@Yuliiadd improvements needed

OleksiyRudenko:
A side note: looks like a PR (code submission) for https://github.com/kottans/frontend/blob/master/tasks/js-pre-oop.md is missing.

Yuliiadd:
@OleksiyRudenko could you give me more hints?)
In first version i used this constructions in parent class method and this worked for all child classes without extending the parent method in them: 

> Object.getOwnPropertyNames(this).map(prop => this[prop]).join('; ')

But after your first comment i thought that i have to change this method extending in child classes.  I did not fully understand your comment, but I changed this method and it turned out that I did not meet the requirements((( I'm confused((  Could you help me to understand whats wrong with this construction? Why i can not use it like in first version?

I can rewrite this method like this, but i don't know this resolve my problem? I will not extend parent method in child class this way: 
```
      let properties = [];
      for(let prop in this){
         properties.push(this[prop]);
      }
      return properties.join('; ');
   }
```

OleksiyRudenko:
`Object.*` methods as well as `for..in` and `for..of` are OK when a particular order of keys/values is not important.
We shouldn't rely on order of class properties definition nor should we change the order of definition when we need to change order of their usage (output in our case).

You are on the right path:

> `` `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.legs}` ``

This is a classic use case for `Array.join`. You also do not need to revert to a good ol' `['name', 'gender'].map(...)`
What we want to eliminate here is repetition of `"; "`. Also if I am tasked to add yet another class property and add it to the output I would prefer to add it twice (in constructor and in output method) without any curly braces etc. I simply can forget to add a space between two properties.

Any valid expression can be an element of an array. Consider the following example:
```
[
  "strings. we already know they are good",
  somePoorlyNamedVariable,
  this.someProperty,
  doSomething(), // a value returned by function will become an array's element
  doSomething, // a function per se; we can call it later
  [], // a nested array
]
```

Your case is simpler then the one above. Hope this helps.

Yuliiadd:
@OleksiyRudenko updated classes methods. 
I hope I understood everything correctly)
Please, re-review)

submissions/YuliiaDyka/oop-exercise/index.js
```diff
@@ -0,0 +1,80 @@
...
+      this.hands = 2;
+   };
+
+   showProperties() {
+      return [
+         super.showProperties(),
+         this.hands,
+      ].join('; ');
+   }
+};
```
OleksiyRudenko, line 68: You may see, how this class breaks OOP, DRY etc
Not your fault.
Here is the learning: OOP is not appropriate for hybrids, non strictly hierarchical relations in the system.
Read about compostion vs inheritance to recognize situations where either is more appropriate.

OleksiyRudenko (approved):
@Yuliiadd great job!

PR #475, opened 2022-09-07
My PR with Tiny Js World

Nik3264:
### Building a Tiny JS World

[Demo](https://nik3264.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Nik3264/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/Nik3264/Building-a-Tiny-JS-World/index.js
+++ b/submissions/Nik3264/Building-a-Tiny-JS-World/index.js
@@ -8,62 +8,65 @@
 // ======== OBJECTS DEFINITIONS ========
 // Define your objects here
 const dog = {
-   species: 'dog',
-   name: 'Toby',
-   gender: 'male',
-   legs: 4,
-   hands: 0,
-   saying: 'woof-woof!'
- };
+  species: "dog",
+  name: "Toby",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+  saying: "woof-woof!",
+};
 
- const cat = {
-   species: 'cat',
-   name: 'Vasya',
-   gender: 'male',
-   legs: 4,
-   hands: 0,
-   saying: 'mioueui-mi-mi!'
- };
+const cat = {
+  species: "cat",
+  name: "Vasya",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+  saying: "mioueui-mi-mi!",
+};
 
- const catwoman = {
-  __proto__: cat, 
-   species: 'catwoman',
-   name: 'Tina',
-   gender: 'male',
-   legs: 4,
-   hands: 0,
-   saying:cat.saying
- };
+const catwoman = {
+  __proto__: cat,
+  species: "catwoman",
+  name: "Tina",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+};
 
- const men = {
-   species: 'men',
-   name: 'Nik',
-   gender: 'male',
-   legs: 2,
-   hands: 2,
-   saying: 'hello!'
- };
+const men = {
+  species: "men",
+  name: "Nik",
+  gender: "male",
+  legs: 2,
+  hands: 2,
+  saying: "hello!",
+};
 
- const woman = {
-   species: 'woman',
-   name: 'Tyna',
-   gender: 'female',
-   legs: 2,
-   hands: 2,
-   saying: 'wo hei! Hello!'
- };
+const woman = {
+  species: "woman",
+  name: "Tyna",
+  gender: "female",
+  legs: 2,
+  hands: 2,
+  saying: "wo hei! Hello!",
+};
 
- const inhabitants = [dog, cat, catwoman, men, woman];
+const properties = ["spesies", "name", "gender", "legs", "hands", "saying"];
+const inhabitants = [dog, cat, catwoman, men, woman];
 
- const listInhabitants = inhabitants.map(inhabitant => {
-    inhabitant.name = `<strong>${inhabitant.name}</strong>`;
+cat.saying = "This is working! Test2!! Is this correct now?";
 
-    return  Object.keys(inhabitant).map(property => {
-        return inhabitant[property];
-    }).join(', ');       
- });
- print(listInhabitants.join(";\n"));
- 
+const listInhabitants = inhabitants.map((inhabitant) => {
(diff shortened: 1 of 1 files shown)


Nik3264:
I have already checked the code, as far as my knowledge was enough.
There are only a few lines here. And this worries me :) I deleted my comments in the code. Only the cats remained ))
Please, review.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,79 @@
...
+   hands: 0,
+   saying: 'mi-mi!'
+ };
+ const catwoman = {
+   species: 'catwoman',
+   name: 'Toby',
+   gender: 'male',
+   legs: 4,
+   hands: 0,
+   saying: cat.saying
```
artem-trubin, line 33: Task asks you to make it so Catwoman always says the same thing the cat says. But in your current implementation, would Catwoman's saying change if we change the cat's one?

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,79 @@
...
+ const woman = {
+   species: 'woman',
+   name: 'Tyna',
+   gender: 'female',
+   legs: 2,
+   hands: 2,
+   saying: 'wo hei!'
+ };
+
+ const arr = [dog, cat, catwoman, men, woman];
```
artem-trubin, line 52: Don't use such broad variable names, especially containing the data type, like `arr`, or `string`. Just describing the content should do the work, for example, this array could be named `inhabitants`. It's plural, which suggests that it is an array and `inhabitant` is pretty defined in your code.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+ const str = arr.map(el=>{
```
artem-trubin, line 54: Don't forget to add spaces around ` => `, ` = ` and pretty much everything. It helps to mare the code clearer.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,79 @@
...
+  el.name=`<strong>${el.name}</strong>`;
+   return Object.values(el).join(', ');
```
artem-trubin, line 56: There is a rule in the task: do not use `Object.values` and similar methods. They are not very reliable in terms of the keys order and presence. Find a way to set a list of keys by hand and work with it.

artem-trubin (changes requested):
Nice work! Let's improve it a little, check the comments below.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: Catwoman.saying’s  [[HomeObject]]  is cat because it was created in cat. The [[HomeObject]] property never changes. And when cat.saying changes -  catwoman.saying also changes, isn’t it? And why in this case it is needed to write __proto__: cat, ?  It is working now without writing it. When I change cat.saying, then catwoman.saying also changes immidiately. Catwoman.saying and cat.saying are interdependent because they are same concepts. We have the result - it works! Is there any issue? I consider it is but I don't know where.  :)  I have alredy read  some info about class inheritance, should I imply this concept here? Could you please give me some hint ? :)  Thank you a lot.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: It looks very easy, so I did not consider it seriously. It is my big mistake! Now I am correcting it.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: Thank you, I correct it.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: I don't know how to do it yet... May I use  Object.keys() ? And then get to values throught  Object.keys() .... ?
Thank you so much.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
artem-trubin: What I mean is this:

```js
const cat = { saying: "First test" };
const catwoman = { saying: cat.saying };
console.log(catwoman.saying) // this will be "First test"
cat.saying = "Second test";
console.log(catwoman.saying) // this should be "Second test", but with your implementation, it will still be "First test"
```

You can use a variety of methods here, prototyping, `bind` etc. Just don't use classes yet.

In this particular task, you are right, it works like this, but the idea is to understand that simple using `saying: cat.saying` just gets the value of `cat.saying` and assigns it once, it is not a link and it will not be updated. And the task asks to make it so it does.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
artem-trubin: > and similar methods

This meant `Object.keys` too. And others like it.

The idea is almost right, you just have to figure out where to get an array of properties to show (hint: make it by yourself)

artem-trubin (changes requested):
I answered some of your comments, so let's consider it a small review. Rerequest another one after you make changes.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: Thank you! I have created an array of properties. now all properties are displayed.

submissions/Nik3264/Building-a-Tiny-JS-World/index.js
Nik3264: Thank you! Now everything is clear. If the property is removed completely, then it is searched for in the prototype. So that's why it was necessary to manually create an array of properties! When using "Object.keys()"  the property "saying" there was none at all. Now I can clearly see the difference.

artem-trubin (approved):
Good job!

PR #477, opened 2022-09-07
a-tiny-js-world

Conversee12:
### a-tiny-js-world

[Demo](https://conversee12.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/Conversee12/a-tiny-JS-world.git)

Please, review.

### DIFF
--- a/submissions/Conversee12/Tiny JS World/index.js
+++ b/submissions/Conversee12/Tiny JS World/index.js
@@ -1,0 +1,49 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/Conversee12/a-tiny-JS-world.git
+   Web app: https://conversee12.github.io/a-tiny-JS-world/
+   */
+
+const Man = {
+   species: 'human',
+   name: 'Vasyl',
+   gender: 'male',
+   legs: 2,
+   hands: 2,
+   saying: 'Nothing can stop an idea whose time has come!',
+};
+
+const Woman = {
+   species: 'human',
+   name: 'Oksana',
+   gender: 'female',
+   legs: 2,
+   hands: 2,
+   saying: 'What\'s up?',
+};
+
+const Dog = {
+   species: 'dog',
+   name: 'Reks',
+   gender: 'male',
+   legs: 4,
+   hands: 0,
+   saying: 'Hav-Hav',
+};
+
+const Cat = {
+   species: 'cat',
+   name: 'Pushok',
+   gender: 'male',
+   legs: 4,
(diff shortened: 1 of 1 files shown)


Conversee12:
I've completed self-checks, did'nt find any common errors

submissions/Conversee12/Tiny JS World/index.js
```diff
@@ -0,0 +1,49 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/Conversee12/a-tiny-JS-world.git
+   Web app: https://conversee12.github.io/a-tiny-JS-world/
+   */
+
+const Man = {
```
artem-trubin, line 8: Usually, names that start with a capital letter describe either class or constructor function. Regular objects (like here) are named with a lowercase first letter.

submissions/Conversee12/Tiny JS World/index.js
```diff
@@ -0,0 +1,49 @@
...
+   species: 'cat',
+   name: 'Pushok',
+   gender: 'male',
+   legs: 4,
+   hands: 0,
+   saying: 'Meow',
+};
+const tinyWorld = [Man, Woman, Dog, Cat];
+
+let newArrayWorld = tinyWorld.map((residents) => residents.species + '; ' + residents.name + '; ' + residents.gender + '; ' + residents.legs + '; ' + residents.hands + '; ' + residents.saying);
```
artem-trubin, line 45: ### Readability

Don't name variables using data types, like `array` or `string`. You just have to choose a name descriptive enough. Lke `inhabitantDescriptions`, for example. You understand what kind of information is there and the plural name suggests that it is an array.

### Pattern
You have a pattern here. Imagine someone would tell you to change the separator from `;` to `,` or something like this. It would be tedious work, especially if your inhabitants would have more properties. 

You have to find a way to specify keys by hand and work with such a list of keys. Pay attention to the task though, you are prohibited to use `Object.values` and methods like these.

submissions/Conversee12/Tiny JS World/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+for (key of newArrayWorld) {
```
artem-trubin, line 47: Every time you have an option to use `Array.forEach` and similar methods instead of `for` and other regular cycles - use them. It increases readability and ensures the reader of your code that every item from the array is operated on, without a need to look into the `for` cycle's setup.

artem-trubin (changes requested):
Nice work! Let's improve it a little. Check the comments below.

artem-trubin (approved):
Good job!

PR #481, opened 2022-09-08
Tiny JS World

JuliaNovak:
### Tiny JS World

[Demo](https://julianovak.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/JuliaNovak/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/JuliaNovak/tiny-js-world/index.js
+++ b/submissions/JuliaNovak/tiny-js-world/index.js
@@ -1,0 +1,76 @@
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+   constructor(species, name, gender, legs, hands, saying) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.legs = legs;
+      this.hands = hands;
+      this.saying = saying;
+      this.properties = [
+         "name",
+         "species",
+         "gender",
+         "legs",
+         "hands",
+         "saying",
+      ];
+   }
+
+   get props() {
+      return this.properties.map((value) => this[value]);
+   }
+
+   display(inhabitantsArr) {
+      inhabitantsArr.map((inhabitant) => {
+         print(inhabitant.props);
+      });
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender);
+      this.legs = 2;
+      this.hands = 2;
+      this.saying = saying;
+   }
+}
+
+class Cat extends Inhabitant {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender);
+      this.legs = 4;
+      this.hands = 0;
+      this.saying = saying;
+   }
+}
+
+class Dog extends Inhabitant {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender);
+      this.legs = 4;
+      this.hands = 0;
+      this.saying = saying;
+   }
+}
+
+class CatWoman extends Cat {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender, saying);
+      this.hands = 2;
+      this.legs = 2;
+   }
+}
+
+const dog = new Dog("dog", "Toby", "male", "Woof!");
+const cat = new Cat("cat", "Lily", "female", "meow!");
+const woman = new Human("woman", "Linda", "female", "What a pretty kitten!");
+const man = new Human("man", "Jim", "male", "Heyyo!");
+const catWoman = new CatWoman("catWoman", "Ro", "female", cat.saying);
+
+// ======== OUTPUT ========
+const inhabitantsArr = [dog, cat, man, woman, catWoman];
+const displayMethod = new Inhabitant();
+displayMethod.display(inhabitantsArr);


JuliaNovak:
have completed self-checks and fixed code accordingly

OleksiyRudenko:
Does this code implemente requirements from [Building a Tiny JS World](https://github.com/kottans/frontend/blob/2022_UA/tasks/js-pre-oop.md) or from [OOP exercise](https://github.com/kottans/frontend/blob/2022_UA/tasks/js-post-oop.md)?

JuliaNovak:
I believe so😅

OleksiyRudenko:
Which of the two?

JuliaNovak:
Oh, sorry
This is from the oop exercise

JuliaNovak:
Hmm, I've done pre-oop TJSW but probably forgot to submit it for review. Gonna fix this now

JuliaNovak:
Should I create a new pull request for the pre-oop TJSW?

OleksiyRudenko:
The rule to follow the instructions is quite universal.

JuliaNovak:
okay, so we can count it as a completed pre-oop task?

JuliaNovak:
or should I upload my old variant of pre-oop and just fix this one to meet all the post-OOP requirements?

OleksiyRudenko:
We need two PRs for two tasks that are different despite based on the same domain area.
It is up to you to decide how you achieve this.
Focus of the review will be different even if you post identical code.
I'd suggest to stick to requirements per each task. This will actually make things easier for the both of us.

JuliaNovak:
Okay, I created a new PR with the pre-oop task. But could you specify which requirements for the post-oop task the code in this PR doesn't meet? 
I checked the requirements and it seems to me that everything is fine, maybe I didn't get something. Thanks!

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+   display(inhabitantsArr) {
+      inhabitantsArr.map((inhabitant) => {
+         print(inhabitant.props);
+      });
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender);
```
OleksiyRudenko, line 34: What `super()` actually is and what it does?
The [docs on the original projetc repo](https://github.com/OleksiyRudenko/a-tiny-JS-world#leveling-your-skills-up) offers 3 articles specifically dedicated to `super`.
Once you find the answer you will fix your code.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+      this.hands = 2;
+      this.saying = saying;
+   }
+}
+
+class Cat extends Inhabitant {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender);
+      this.legs = 4;
+      this.hands = 0;
```
OleksiyRudenko, line 45: OOP helps us to skip properties that objects do not need.
Imagine we have a complete bio taxonomy. Making every living being have properties that they do not have by natural design clutters them and forces us to handle special cases.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+      super(species, name, gender, saying);
+      this.hands = 2;
+      this.legs = 2;
+   }
+}
+
+const dog = new Dog("dog", "Toby", "male", "Woof!");
+const cat = new Cat("cat", "Lily", "female", "meow!");
+const woman = new Human("woman", "Linda", "female", "What a pretty kitten!");
+const man = new Human("man", "Jim", "male", "Heyyo!");
```
OleksiyRudenko, line 70: Men and women are the same species.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+class CatWoman extends Cat {
+   constructor(species, name, gender, saying) {
+      super(species, name, gender, saying);
+      this.hands = 2;
+      this.legs = 2;
+   }
+}
+
+const dog = new Dog("dog", "Toby", "male", "Woof!");
+const cat = new Cat("cat", "Lily", "female", "meow!");
```
OleksiyRudenko, line 68: Why do we want a user to specify species for dogs and cats?
This property can be "silently" assigned by classes.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+const woman = new Human("woman", "Linda", "female", "What a pretty kitten!");
+const man = new Human("man", "Jim", "male", "Heyyo!");
+const catWoman = new CatWoman("catWoman", "Ro", "female", cat.saying);
+
+// ======== OUTPUT ========
+const inhabitantsArr = [dog, cat, man, woman, catWoman];
+const displayMethod = new Inhabitant();
```
OleksiyRudenko, line 75: Reads weird. We create new Inhabitant and name it displayMethod.
Every instance inherits properties and methods from base classes that we want to use/re-use.
Check OOP theory materials.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+   get props() {
+      return this.properties.map((value) => this[value]);
+   }
+
+   display(inhabitantsArr) {
+      inhabitantsArr.map((inhabitant) => {
+         print(inhabitant.props);
+      });
+   }
```
OleksiyRudenko, line 29: A methopd should handle the class instance's properties.

OleksiyRudenko (changes requested):
@JuliaNovak nice try.
OOP requires us to change the approach and it also helps to write expressive and DRY code.
Think of 1000's of inhabitants with 10's of varying properties.
pre-OOP paradigm would make it a nightmare to maintain a big world. OOP is to the rescue.

JuliaNovak:
still working on this pr

submissions/JuliaNovak/tiny-js-world/index.js
JuliaNovak: Thanks for the links!
The problem is that I've read these articles a couple of times but I still don't see the mistake. Maybe u could point to it or give some more hints?
Answering the question, I think that super here acts as a function and it references the variables in the parent class

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,70 @@
...
+      this.saying = saying;
+   }
+
+   display() {
+      console.log(this);
+      let values = "";
+      for (const prop in this) {
+         values += ` ${this[prop]}`;
+      }
+      print(values);
```
OleksiyRudenko, line 19: Use your learnings from TJSW and Frogger to improve code here.

submissions/JuliaNovak/tiny-js-world/index.js
```diff
@@ -0,0 +1,70 @@
...
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying) {
+      super("human", name, gender);
```
OleksiyRudenko, line 25: What `super` is and what it does? How did you decide to use it this way?
This is not a rhetorical question.

OleksiyRudenko (changes requested):
@JuliaNovak use your learnings from past tasks and OOP modules.

submissions/JuliaNovak/tiny-js-world/index.js
JuliaNovak: I've answered this question above: "Thanks for the links! The problem is that I've read these articles a couple of times but I still don't see the mistake. Maybe u could point to it or give some more hints? Answering the question, I think that super here acts as a function and it references the variables in the parent class"
I've read that there are two ways of using _super_: the one is represented in the code snippet above and another one is super.methodName(). I suppose that the first variant is used to reference the variables in the parent class and the second one - to reference the methods in the parent class. Since I need variables, I chose the 1st variant

OleksiyRudenko: What happens when `super` is called in the child class' constructor? What is called actually?

JuliaNovak: it calls the parent's constructor as I understand

JuliaNovak: may I also say that it calls the parent's prototype? or is it not correct?

PR #483, opened 2022-09-08
A Tiny JS World

iva-stasia:
### A Tiny JS World

[Demo](https://iva-stasia.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/iva-stasia/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/iva-stasia/tiny-js-world/index.js
+++ b/submissions/iva-stasia/tiny-js-world/index.js
@@ -18,7 +18,7 @@ const cat = {
    saying: 'Meeeow!',
 };
 
-const women = {
+const woman = {
    species: 'human',
    name: 'Gwen',
    gender: 'female',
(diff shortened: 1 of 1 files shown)


iva-stasia:
Checked my code. It looks like there are no common mistakes.

iva-stasia:
I realized that I can make my code shorter and got rid of the extra array.

submissions/iva-stasia/tiny-js-world/index.js
```diff
@@ -0,0 +1,51 @@
...
+   saying: 'What\'s up?'
+};
+
+const catWomen = {
+   species: 'cat-human',
+   name: 'Dita',
+   gender: 'female',
+   legs: 2,
+   hands: 2,
+   saying: cat.saying
```
artem-trubin, line 45: Your catwoman's saying should change when cat's saying changes. Your's does not. You only take the initial value of the `cat.saying`.

artem-trubin (changes requested):
Nice work, you're almost there. Let's fix something, check the comment below.

submissions/iva-stasia/tiny-js-world/index.js
iva-stasia: Now I understand that it only works once when I create the object.
Changed it.

artem-trubin (approved):
Good job!

PR #491, opened 2022-09-08
Add object-oriented-js task

misshatsker:
### Object oriented JS

[Demo](https://misshatsker.github.io/frontend-nanodegree-arcade-game) |
[Code base](https://github.com/misshatsker/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/misshatsker/object-oriented-js/app.js
+++ b/submissions/misshatsker/object-oriented-js/app.js
@@ -1,51 +1,71 @@
-var leftSide = 0;
-var rightSide = 500;
-var topSide = 0;
-var bottomSide = 500;
-var initialGroundY = 60;
+const leftSide = 0;
+const topSide = 0;
+const rightSide = 500;
+const bottomSide = 500;
 
-var Enemy = function(x, y, speed) {
-    this.sprite = 'images/enemy-bug.png';
+const initialGroundY = 60;
+const tileHeight = 80;
+
+const enemySprite = 'images/enemy-bug.png';
+const enemyWidth = 80;
+const enemyHeight = 70;
+const enemyMinSpeed = 100;
+const enemyMaxSpeed = 300;
+const enemyMinX = -rightSide;
+const enemyMaxX = enemyWidth - rightSide;
+
+const playerSprite = 'images/char-princess-girl.png';
+const playerWidth = 70;
+const playerHeight = 70;
+const playerSpeed = 75;
+const playerInitialX = 200;
+const playerInitialY = 400;
+
+const Character = function(x, y, speed, sprite) {
+    this.sprite = sprite;
     this.speed = speed;
     this.x = x;
     this.y = y;
 };
 
+Character.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const Enemy = function(...args) {
+    Character.call(this, ...args);
+};
+
 Enemy.prototype.update = function(dt) {
-    var shift = this.speed * dt;
+    const shift = this.speed * dt;
     this.x = this.x + shift;
     if (this.x >= rightSide) {
         this.x = -this.width;
     }
 };
 
-Enemy.prototype.render = function() {
-    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+Enemy.prototype.width = enemyWidth;
+Enemy.prototype.height = enemyHeight;
+
+Object.setPrototypeOf(Enemy.prototype, Character.prototype);
+
+const Player = function(...args) {
+    Character.call(this, ...args);
 };
 
-Enemy.prototype.width = 80;
-Enemy.prototype.height = 70;
-
-var Player = function(x, y, speed) {
-    this.sprite = 'images/char-princess-girl.png';
-    this.speed = speed;
-    this.x = x;
-    this.y = y;
-};
-
-Player.prototype.width = 70;
-Player.prototype.height = 70;
+Player.prototype.width = playerWidth;
+Player.prototype.height = playerHeight;
 
 Player.prototype.update = function() {
     const isOverlapDetected = allEnemies.some((enemy) => {
         return checkIsOverlap(
             {
-                x: player.x,
-                y: player.y
+                x: this.x,
+                y: this.y
             },
             {
-                x: player.x + player.width,
-                y: player.y + player.height
+                x: this.x + this.width,
+                y: this.y + this.height
             },
             {
                 x: enemy.x,
@@ -61,10 +81,6 @@ Player.prototype.update = function() {
     if (isOverlapDetected) {
         gameOver(false);
     }
-}
-
-Player.prototype.render = function() {
-    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
 }
 
 Player.prototype.handleInput = function(action) {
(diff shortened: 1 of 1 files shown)


misshatsker:
> Hey!
> 
> Congratulations on your PR! 😎😎😎
> 
> Let's do some **self-checks** to fix most common issues and to make some improvements to the code before reviewers put their hands on the code.
> 
> Go through the requirements/most common mistakes linked below and fix the code as appropriate.
> 
> If you have any questions to requirements/common mistakes **feel free asking** them here or in Students' chat.
> 
> When you genuinely believe you are done **put a comment** stating that you have completed self-checks and fixed code accordingly.
> 
> Also, be aware, that if you would silently ignore this recommendation, a mentor can think that you are still working on fixes. And your PR will not be reviewed. 😒
> 
> Please, make sure that your code follows [the requirements](https://kottans.org/documentation/docs/doc/code-review/#oo-js-frogger)
> 
> Universal recommendations:
> 
> * Make sure your code follows [General Requirements](https://kottans.org/documentation/docs/doc/code-review/#general-requirements)
> * Use common sense or seek for an advice whenever requirements look ambiguous or unclear.
> 
> By the way, you may proceed to the next task before this one is reviewed and merged.
> 
> _Sincerely yours,_ _Submissions Kottachecker_ 😺

I have finished self-check

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,171 @@
...
+Player.prototype.update = function() {
+    const isOverlapDetected = allEnemies.some((enemy) => {
+        return checkIsOverlap(
+            {
+                x: player.x,
+                y: player.y
+            },
+            {
+                x: player.x + player.width,
+                y: player.y + player.height
```
OleksiyRudenko, line 48: Why we refer to a specific `Player` instance inside class definition?
What if we have two players and rename that global variable?
Make the code safe.
Also requirements address this situation specifically.
Please find it, post here as a comment and fix code accordingly.

Also check other fragments of the code against this requirement and fix accordingly.

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,171 @@
...
+var allEnemies = [
+    createRandomEnemy(0),
+    createRandomEnemy(1), 
+    createRandomEnemy(2)
+];
+
+var initialPlayerX = 200;
+var initialPlayerY = 400;
+var playerSpeed = 75;
+var player = new Player(initialPlayerX, initialPlayerY, playerSpeed);
```
OleksiyRudenko, line 127: Requirements address issue we have in this fragment and in many places across the code.
Please find it, post here as a comment and fix code accordingly.

OleksiyRudenko (changes requested):
@misshatsker 
Please copy requirements posted/linked by bot as a comment in this PR.
Go through every single one and mark those you are genuinely sure are met.
If any remain unclear, feel free asking specific questions here and/or Students' chat.

Also see comments below.

misshatsker:
@OleksiyRudenko I'm so sorry for requesting a review without proper self-checking. Please check the marked list of requirements and changes I've made.

- [x] Employ ES6 features like const, let etc. (with exclusion of ES6 class syntax).
- [x] The code is very [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
- [x] Requirements regarding Constants:
  - [x] All numbers like block dimensions, initial locations are defined as constants.
  - [x] Every number that has a semantic purpose (like those listed above) should be defined as constants; think of how your code reads - the closer to plain English the better.
  - [x] There are core constants and derived (calculated) constants (e.g. const FIELD_WIDTH = BLOCK_WIDTH * BLOCKS_NUMBER;).
  - [x]  Arrays of constants also qualify as constants, e.g. const INITIAL_POSITIONS = [1,2,3,4].map(rowNumber => rowNumber * BLOCK_HEIGHT);
  - [ ] Const objects help organizing and structure const data even better, e.g. const PLAYER_CONF = { initialPosition: {x: 1, y: 5}, sprite: '...', ...etc... };
- [x] Requirements regarding OOP:
  - [x] Properties common for some classes are generalized into a base class (e.g. there is Character base class, which is extended by Enemy and Player classes).
  - [ ] Class extension is implemented using Subclass.prototype = Object.create(Superclass.prototype), not Subclass.prototype = new Superclass(params);; [Useful resource](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain). Note: I've used `Object.setPrototypeOf(..)` here. Is it legitimate?
  - [x] Classes do not refer to any global variables, like global variable player, which is an instance of Player class (referring to global constants and globals provided by the gaming platform like Resources is OK); Hint: pass instance of a game object (or objects) as an argument to other game objects they need to interact with.
  - [x] Separation of Concerns principle is followed (e.g. update method only updates character coordinates and doesn't contain any inline code to check e.g. collisions; calling other methods from update is legitimate).
- [x] Most common mistakes

submissions/misshatsker/object-oriented-js/app.js
misshatsker: My fault. I should have used `const`s/`let`s instead of `var`s. Not for an excuse but I've been misled by using var in the original repo.

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,188 @@
...
+    this.speed = speed;
+    this.x = x;
+    this.y = y;
+};
+
+Character.prototype.render = function() {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+const Enemy = function(...args) {
```
OleksiyRudenko, line 35: Bad practice.
IDE intellisense won't be able to hint what arguments are expected.
Is there any good reason to have `...args`?
Why not using named parameters?

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,188 @@
...
+Enemy.prototype.update = function(dt) {
+    const shift = this.speed * dt;
+    this.x = this.x + shift;
+    if (this.x >= rightSide) {
+        this.x = -this.width;
+    }
+};
+
+Enemy.prototype.width = enemyWidth;
+Enemy.prototype.height = enemyHeight;
```
OleksiyRudenko, line 48: Why mixed style of assigning class properties?

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,188 @@
...
+    const y = initialGroundY + row * tileHeight;
+    const speed = getRandomNumberFromARange(enemyMinSpeed, enemyMaxSpeed);
+
+    return new Enemy(x, y, speed, enemySprite);
+}
+
+const allEnemies = [
+    createRandomEnemy(0),
+    createRandomEnemy(1), 
+    createRandomEnemy(2)
```
OleksiyRudenko, line 118: Imagine, you are tasked to add 10 more enemies. Make this fragment DRY.

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,188 @@
...
+
+    const isOneRectangleAboveAnother = (r1.y < l2.y || r2.y < l1.y);
+    if (isOneRectangleAboveAnother) {
+        return false;
+    }
+
+    return true;
+}
+
+function checkPlayerOverlapWithEnemies(player) {
```
OleksiyRudenko, line 148: Why not a part of Enemy's or Player's behaviour?
We could build the entire game on global functions.
There is a reason for OOP to exist.

OleksiyRudenko (changes requested):
@misshatsker good start.
Needs improvement.

submissions/misshatsker/object-oriented-js/app.js
misshatsker: Just to not copy/paste them as far as I wanted to reuse the `Character` constructor completely. But I got your notice with IDE, thanks! I haven't thought about that

submissions/misshatsker/object-oriented-js/app.js
misshatsker: You're right, they shouldn't be there. I considered them as common properties, but they aren't

submissions/misshatsker/object-oriented-js/app.js
misshatsker: I've update it, check it please

submissions/misshatsker/object-oriented-js/app.js
```diff
@@ -0,0 +1,194 @@
...
+function createRandomEnemy(row) {
+    const x = getRandomNumberFromARange(enemyMinX, enemyMaxX);
+    const y = initialGroundY + row * tileHeight;
+    const speed = getRandomNumberFromARange(enemyMinSpeed, enemyMaxSpeed);
+
+    return new Enemy(x, y, speed, enemySprite, enemyWidth, enemyHeight);
+}
+
+function createRandomEnemies(amount) {
+    return new Array(amount).fill(null).map((elem, i) => createRandomEnemy(i));
```
OleksiyRudenko, line 136: There is a convention to use `_` for the `map` argument when it is not actually used, so `...map((_, i) => ...)`

OleksiyRudenko (approved):
@misshatsker great job done!

PR #518, opened 2022-09-11
OOP Exercise

unabyband:
### OOP Exercise

[Demo](https://unabyband.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/unabyband/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/unabyband/oop_exercise/index.js
+++ b/submissions/unabyband/oop_exercise/index.js
@@ -21,6 +21,7 @@ class Inhabitant {
   printInhabitant() {
     const nameToPrint = `<strong>${this.name}</strong>`;
     const talkToPrint = `<em>${this.saying}</em>`;
+    const legsToPrint = `On ${this.legs} legs`;
     const friendsToPrint = this.friends
       ? this.friends.map((friend) => friend.name).join(", ")
       : `${this.species} has no friends!`;
@@ -30,7 +31,7 @@ class Inhabitant {
         this.species,
         nameToPrint,
         this.gender,
-        this.legs,
+        legsToPrint,
         talkToPrint,
         friendsToPrint,
       ].join("; ")
@@ -40,15 +41,37 @@ class Inhabitant {
 
 class Human extends Inhabitant {
   constructor(name, gender, saying, friends) {
-    const [legs, hands] = [`On 2 legs`, `Has 2 hands`];
-    super("human", name, gender, [legs, hands].join("; "), saying, friends);
+    const legs = 2;
+    const hands = 2;
+    super("human", name, gender, legs, saying, friends);
+    this.hands = hands;
+  }
+
+  //Rewriting the parent class method for printing out the string which includes unique property 'hands' for Human
+  printInhabitant() {
+    const nameToPrint = `<strong>${this.name}</strong>`;
+    const legsToPrint = `On ${this.legs} legs`;
+    const handsToPrint = `Has ${this.hands} hands`;
+    const talkToPrint = `<em>${this.saying}</em>`;
+    const friendsToPrint = this.friends.map((friend) => friend.name).join(", ");
+
+    print(
+      [
+        this.species,
+        nameToPrint,
+        this.gender,
+        legsToPrint,
+        handsToPrint,
+        talkToPrint,
+        friendsToPrint,
+      ].join("; ")
+    );
   }
 }
 
 class Animal extends Inhabitant {
   constructor(species, name, gender, saying, friends) {
-    const paws = 4;
-    super(species, name, gender, `On ${paws} paws`, saying, friends);
+    super(species, name, gender, 4, saying, friends);
   }
 }
 
--- a/submissions/unabyband/oop_exercise/index.js
+++ b/submissions/unabyband/oop_exercise/index.js
@@ -9,29 +9,30 @@
 // Define your objects here
 
 class Inhabitant {
-  constructor (species, name, gender, limbs, saying, friends) {
+  constructor (species, name, gender, legs, hands, saying, friends) {
     this.species = species;
-    this.name = name;
+    this.name = `<strong>${name}</strong>`;
     this.gender = gender;
-    this.limbs = limbs;
-    this.saying = saying;
+    this.legs = legs;
+    this.hands = hands;
+    this.saying = `<em>${saying}</em>`;
     this.friends = friends;
   }
 
   printInhabitant () {
-    print(`${this.species}; <strong>${this.name}</strong>; ${this.gender}; ${this.limbs.join('; ')}; <em>${this.saying}</em>; ${this.friends ? this.friends.map(elem => elem.name).join(', ') : 'Cats have no friends'}`);
+    print((Object.values(this)).join("; "));
   }
 }
 
 class Human extends Inhabitant {
   constructor (name, gender, saying, friends) {
-    super('human', name, gender, [2, 2], saying, friends);
+    super('human', name, gender, 2, 2, saying, friends);
   }
 }
 
 class Animal extends Inhabitant {
   constructor (name, gender, saying, friends) {
-    super('animal', name, gender, [4], saying, friends);
+    super('animal', name, gender, 4, 0, saying, friends);
   }
 }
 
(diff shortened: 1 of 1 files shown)


unabyband:
I've made self-checking, please review

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+class Inhabitant {
+  constructor (species, name, gender, limbs, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.limbs = limbs;
```
OleksiyRudenko, line 16: Introduce legs and hands and/or paws. In OOP-compliant manner.
Will be happy to see if you mastered OOP - that's the purpose of the module.

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+  new Dog('Rex', 'male'),
+  new Dog('Lizzy', 'female') 
+ ];
+ 
+ /*Here's my function for adding friends to inhabitants, 
+ the first argument is index of subject at inhabitants array
+ all next arguments are indexes of his friends.
+ I know, it's optional, but hoping my solution is not too primitive*/
+
+ function addFriends(...arguments) {
```
OleksiyRudenko, line 68: With current signature won't let me know what exactly is expected by this function at call time, without reading the source code.
Code is written for other people.

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+ }
+
+addFriends(0, 3, 7);
+addFriends(1, 4, 5, 6);
+addFriends(2, 5, 6, 9);
+addFriends(3, 0, 7);
+addFriends(4, 1, 5, 6, 8);
+addFriends(5, 2, 6, 9);
+addFriends(8, 4);
+addFriends(9, 1, 2);
```
OleksiyRudenko, line 83: The function signature looked like all arguments represent the same role. This example doesn't really give an idea of how to add another friend pack. Are these numbers a sort of ids?
Make it more obvious

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,87 @@
...
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.limbs = limbs;
+    this.saying = saying;
+    this.friends = friends;
+  }
+
+  printInhabitant () {
+    print(`${this.species}; <strong>${this.name}</strong>; ${this.gender}; ${this.limbs.join('; ')}; <em>${this.saying}</em>; ${this.friends ? this.friends.map(elem => elem.name).join(', ') : 'Cats have no friends'}`);
```
OleksiyRudenko, line 22: There are `Array` methods to make code DRYer. An array element can be an expression (or rather a result of expression).

OleksiyRudenko (changes requested):
@unabyband one of the purposes of the module is to learn something new and also demonstrate a practical grip over this new knowledge.
Also techniques learned earlier still worth re-using.

submissions/unabyband/oop_exercise/index.js
unabyband: Done!

submissions/unabyband/oop_exercise/index.js
unabyband: Done!

submissions/unabyband/oop_exercise/index.js
unabyband: Well, I finally give up. 
I've tried so many variations how to force this function works as a method of "Inhabitant" class, but it doesn't look possible before I initialize "inhabitants" array.
Perhaps, this function should not be the method of main class. I don't know
Or, maybe lot easier resolve exists and it just unclear for me at this point.
Sorry, I remember that this part of exercise does not necessary
We can put whole objects as elements of each "friends" container only after the initialization of our array not before.

OleksiyRudenko: You can call any method after the object initialization. E.g. `paul.addFriends([nina, toby, robert])`

submissions/unabyband/oop_exercise/index.js
OleksiyRudenko: `Object` methods still do not guarantee order of properties/values.
You could improve exisiting line 22.

submissions/unabyband/oop_exercise/index.js
unabyband: But in this case, I should totally change the construction of our array, in order the name of inhabitant becomes the name of variable, I guess...

submissions/unabyband/oop_exercise/index.js
unabyband: Do I need deconstruct this method?

OleksiyRudenko: Not really. We speak about `[].join` and
> An array element can be an expression (or rather a result of expression).

Feel free to ask questions

unabyband: But our inhabitants are objects, not the arrays, do we need to convert them or use some sort of map method?
I'm not sure if this will make my code DRYer...

unabyband: Ok, I think I finally get it, please review again

unabyband:
Please review again, thank you

submissions/unabyband/oop_exercise/index.js
unabyband: Could you review it again, please?

submissions/unabyband/oop_exercise/index.js
OleksiyRudenko: You can extract names (or any other property) when it is needed exactly.
Consider the following example:
```
function setParents(parents) { this.parents = parents }
function getParentsNames() { return this.parents.map(parent => parent.name).join(", ") }
```
Rule of thumb: extract and transform original data exactly where it is needed.

OleksiyRudenko:
Bring back paws and hands (in OOP way), friends and whatever you just removed from the code.
And try to solve this problems. You will learn something useful that you will be able use in your next projects. You will also understand better certain important concepts on Stage 1.
Do not worry about stage 0 deadline.

OleksiyRudenko (changes requested):
@unabyband 
Bring back paws and hands (in OOP way), friends and whatever you just removed from the code.
And try to solve this problems. You will learn something useful that you will be able use in your next projects. You will also understand better certain important concepts on Stage 1.
Do not worry about stage 0 deadline.

unabyband:
I bring back and rewritten addFriends function, and make some fixes, please re-review. Thank you

unabyband:
I've brought legs/hands/paws back, and made some out formatting for greater look at the page.
Also I've wrote (and modified after that) addFriends() function and used "Prettier" for better code look.
Please re-review my work, thanks a lot!

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    super("human", name, gender, 2, 2, saying, friends);
+  }
+}
+
+class Animal extends Inhabitant {
```
OleksiyRudenko, line 52: In fact, every instance of `Animal` has property for hands - `hands`.
OOP helps to design the system of business entities so that they just do not have unnecessary properties.
This is something bigger than just a way to provide default values to the base class.

Also requirement say that inhabitants that do not have something by natural design they do not have any properties for this. OOP helps to build scalable code that doesn't require to change base class when we want to handle something specific to a sub-class.

Make so that only humans have hands in any meaning. Also remember that very class should handle the properties it owns immediately. Do not delegate property handling to parent class (it should be agnostic of any properties a child class may have).

OleksiyRudenko (changes requested):
@unabyband good improvements.
Now need to make the design OOP-compliant.

submissions/unabyband/oop_exercise/index.js
unabyband: Well, I've remade legs/hands/paws properties for ancestor and child classes. Now the Animal class doesn't have any type of 'hands' property.
For best (in my opinion) improvement - I decided to pass into legs property of ancestor class already finished string contains formatted out.
Though, I also thought about rewriting printInhabitant method especially for Human class, just in order to printing hands and legs unlike the named method in ancestor class, but considered this way less DRY (Even though rewriting methods is one of the OOP-principles). 
Please re-review my fixes. Thank you.
And please, excuse me for long feedback time, I had to finished current work.

unabyband:
Now Base class and Animal subclass don't have any type of "Hands" property, this became exclusively for Human.
Please re-review

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+  constructor(species, name, gender, legs, saying, friends) {
+    this.species = species;
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this.saying = saying;
+    this.friends = friends;
+  }
+
+  printInhabitant() {
```
OleksiyRudenko, line 21: (1) Separation of concerns (remember requirements re Frogger?)
This method forms an output and also conducts output.

(2) You may construct arrays from any legal JS expressions. Consider an example below
```
[
  this.propName,
  `prefix ${this.PropName} suffix`,
  arrayVariable.join("glue "),
  condition ? expressionIfTrue : "stringIfFalse",
]

```

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,122 @@
...
+    this.hands = hands;
+  }
+
+  //Rewriting the parent class method for printing out the string which includes unique property 'hands' for Human
+  printInhabitant() {
+    const nameToPrint = `<strong>${this.name}</strong>`;
+    const legsToPrint = `On ${this.legs} legs`;
+    const handsToPrint = `Has ${this.hands} hands`;
+    const talkToPrint = `<em>${this.saying}</em>`;
+    const friendsToPrint = this.friends.map((friend) => friend.name).join(", ");
```
OleksiyRudenko, line 56: Use method overloading/overriding. This will help to make your code DRY.
Check learning materials and find out what `super`'s extended API can do for you.

OleksiyRudenko (changes requested):
@unabyband quite a decent job.
Let's make it OOP compliant and use mechanics JS OOP syntax offers to write DRY, maintainable and scalable code.

unabyband:
Thanks for your review (especially for the hint by extended using "super"). 
I've separated method printInhabitant in order to make DRY-er code, and also use some contractions for made the out string/array of the inhabitants (in fact I've already used that type of array construction in one of the previous commits, but wasn't sure if it correct in this exercise).
Please re-review it again, many thanks

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,109 @@
...
+class Human extends Inhabitant {
+  constructor(name, gender, saying, friends) {
+    const legs = 2;
+    const hands = 2;
+    super("human", name, gender, legs, saying, friends);
+    this.hands = hands;
+  }
+
+  printInhabitant() {
+    super.printToPage(this.species,
```
OleksiyRudenko, line 50: Why not re-use parent's printInhabitant and then adding what is specific to this class?
I see, because parent's `printInhabitant` is not really re-usable as it returns nothing.
Use `print` function completely outside of classes definition.

OleksiyRudenko (changes requested):
@unabyband almost there

submissions/unabyband/oop_exercise/index.js
unabyband: I've made it, please review again

submissions/unabyband/oop_exercise/index.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  printInhabitant() {
+    return [this.species,
+        `<strong>${this.name}</strong>`,
+        this.gender,
+        `On ${this.legs} legs`,
+        `Has ${this.hands} hands`,
+        `<em>${this.saying}</em>`,
+        this.friends.map((friend) => friend.name).join(", ")].join('; ');
```
OleksiyRudenko, line 50: Now re-use parent's version of the method and add whatever this class needs to add.
Let's have hands at the end for the sake of simplicity and in favour of OOP compliance.

Alternative would be to keep properties as array all way down until you need to print them. This way you could insert hands somewhere after legs or elsewhere. But this too much change. Please just keep this in mind. And attach hands at the end of the parent's method output.

unabyband: I've made it, please review. And sorry for delayed answer, because of missile attacks we don't have an electricity and cable internet.
Thank you

OleksiyRudenko (approved):
@unabyband great job!

PR #530, opened 2022-09-12
Oop task

sofiichuk:
### OOP exercise

[Demo](https://rawcdn.githack.com/sofiichuk/a-tiny-JS-world/52fb970af4d58fe3385f9c435a4e67dc61c467c3/index.html) |
[Code base](https://github.com/sofiichuk/a-tiny-JS-world/blob/OOP-exercise/index.js)

Please, review.


sofiichuk:
i checked it.
i have doubdts 'bout this, cause i don't really get it: "There is no need to specify species at instantiation yet species are printed".
but i hope it's not compulsory.
also i decided it's better to delete arms, not to add - and i think the logic demands it.
and i guees, the rest is correct

OleksiyRudenko:
> i checked it. i have doubdts 'bout this, cause i don't really get it: "There is no need to specify species at instantiation yet species are printed". 

The following would work:
```
const kitten = new Cat("Pinkie", "female", 4);
print(kitten) // cat, Pinkie, female, 4
```
Focus here on species. Other props are just for the context.

> but i hope it's rather a recommendation than a requirement
Answer is: code has to be DRY. And DRY is a requirement. `new Cat("cat"); new Human("human"); new Dog("dog")` is not DRY enough.

> also i decided it's better to delete arms, not to add - and i think the logic demands it. and i guees, the rest is correct
We add all sorts of limbs (paws, legs, hands, wings, fins) and then remove all but required. Is it really logical?

submissions/sofiichuk/oop-exercise/index.js
```diff
@@ -0,0 +1,65 @@
...
+        super(name, gender, saying);
+        this.legs = legs;
+        this.arms = arms;
+    }
+}
+
+class Pet extends Animal {
+    constructor(name, gender, saying, legs = 4) {
+        super(name, gender, saying);
+        this.legs = legs;
```
OleksiyRudenko, line 20: What is so special about `legs` in comparison with e.g. `name` so that it is not generalized into the base class?

submissions/sofiichuk/oop-exercise/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+const Buddy = new Dog('Buddy', 'male');
+const Betty = new Dog('Betty', 'female');
+
+const Kitty = new Cat('Kitty', 'female');
+const Tiger = new Cat('Tiger', 'male');
+
+const creatures = [Tom, Dick, Harry, Polly, Molly, Mary, Buddy, Betty, Kitty, Tiger];
+const creaturePropsList = creatures.map(props => `${Object.values(props).join(";")}`)
+    .join('\n\n');
```
OleksiyRudenko, line 64: (1) The floowing requirement is still valid:
> Object methods like keys, values, entries shouldn't be used when a particular order is required as these do not guarantee any particular order of keys/values. Same refers to for...of and for...in when applied to objects. Hint: List explicitly the properties used to form an object presentation string.

(2) OOP also allows to encapsulate behaviour (as methods), where every class handles properties it owns.

OleksiyRudenko (changes requested):
@sofiichuk improvements are needed.

Why no species?

> There is no need to specify species at instantiation yet species are printed

Look, how you do not specify legs and/or hands at objects instantiation but they are still printed.

submissions/sofiichuk/oop-exercise/index.js
sofiichuk: moved them to the base class

submissions/sofiichuk/oop-exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class Human extends Animal {
+    constructor(name, gender, saying, legs = 2, arms = 2) {
+        super(name, gender, saying, legs);
+        this.arms = arms;
+    }
+
+    declareGenus() {
+        return [this.constructor.name, super.declareAnimal(), this.arms].join(";");
```
OleksiyRudenko, line 21: Check what is the output of `this.constructor.name` when it is used in the root class.
You will possibly reduce code repetition.
BTW, there was no requirement to have classes for men and women, just for species.
I think dogs and cat may feel discriminated, as they do not have privilege of gender-related classes like `HeDog/SheDog` and `HeCat/SheCat`.

submissions/sofiichuk/oop-exercise/index.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class Pet extends Animal {
+    constructor(name, gender, saying, legs = 4) {
+        super(name, gender, saying, legs);
+    }
+
+    declareGenus() {
+        return [this.constructor.name, super.declareAnimal()].join(";");
+    }
```
OleksiyRudenko, line 32: `declareGenus` rather represents class instance.
It is OK to give the same name to the methods that have basically the same purpose in child classes.
`super.methodName` will call inherited version anyway.

OleksiyRudenko (changes requested):
@sofiichuk a pretty decent job done!
Let's polish it.

submissions/sofiichuk/oop-exercise/index.js
sofiichuk: hi.
i will fix what you're asking for considering the code, but there's another point. 
you ask to set classNames/classHierarchy properly. 
no way, the exercise is initially broken. 
there's a general trouble with class naming regarding this task. 
i've gotta philosophical education (and also i'm a pedant) (and it spoils everything).
i need man-women classes for the purpose of some symmetry.
due to aristotelis, we define anything by naming its genus+sigificant_feature (walnut is a tree with nuts). 
the base class should be the god, but let's skip it, it's a joke. 
definitely, the base-js-class is an animal. but if we dig deeper into the taxonomy - levels of human & cats-dogs (canivals) are not equal. parets-childs has a bunch of levels between also. and so on, and so on. what i'm trying to say: ANY approach to naming in this certain task is falthy. the only thing we can do - minimizing stupidity of simpliefied approach

sofiichuk: i mean we write a code not only for the machines, but also for other people. 
(the text above is a bit messy, but nevertheless)

sofiichuk: removed redundant classes. and also species printing is now defined by the root class

OleksiyRudenko (approved):
@sofiichuk great job done!

PR #540, opened 2022-09-13
Add OOP Tiny JS World task

AriadnaKravchuk:
### OOP Tiny JS World

[Demo](https://ariadnakravchuk.github.io/a-tiny-JS-world) |
[Code base](https://github.com/AriadnaKravchuk/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
+++ b/submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
@@ -1,0 +1,111 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: https://github.com/AriadnaKravchuk/a-tiny-JS-world
+   Web app: https://ariadnakravchuk.github.io/a-tiny-JS-world
+*/
+
+// ======== OBJECTS DEFINITIONS ========
+
+const SPECIES = {
+   HUMAN: "human",
+   CAT: "cat",
+   DOG: "dog"
+}
+
+const GENDER = {
+   MALE: "male",
+   FEMALE: "female"
+}
+
+class Inhabitant {
+   constructor(species, name, gender, saying) {
+      this.species = species;
+      this.name = name;
+      this.gender = gender;
+      this.saying = saying;
+      this.friends = [];
+   }
+
+   addFriends(...friend) {
+      if (!this.friends.includes(friend)) this.friends.push(friend);
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(species, name, gender, saying, legs, hands) {
+      super(species, name, gender, saying);
+      this.legs = legs;
+      this.hands = hands;
+   }
+
+   printAttributes() {
+      print([
+         this.species,
+         this.name,
+         this.gender,
+         this.saying,
+         this.legs,
+         this.hands
+      ].join("; "));
+   }
+}
+
+class Animal extends Inhabitant {
+   constructor(species, name, gender, saying, paws) {
+      super(species, name, gender, saying);
+      this.paws = paws;
+   }
+   
+   printAttributes() {
+      print([
+         this.species,
+         this.name, 
+         this.gender, 
+         this.saying, 
+         this.paws
+      ].join("; ")); 
+   }
+}
+
+class Dog extends Animal {
+   constructor(name, gender, saying, paws) {
+      super(SPECIES.DOG, name, gender, saying, paws);
+   }
+}
+
+class Cat extends Animal {
+   constructor(name, gender, saying, paws) {
+      super(SPECIES.CAT, name, gender, saying, paws);
+   }
+}
+
+class Man extends Human {
+   constructor(name, saying, legs, hands) {
+      super(SPECIES.HUMAN, name, GENDER.MALE, saying, legs, hands);
+   }
+}
+
+class Woman extends Human {
+   constructor(name, saying, legs, hands) {
+      super(SPECIES.HUMAN, name, GENDER.FIMALE, saying, legs, hands);
+   }
+}
+
+const dog = new Dog("Pie", GENDER.MALE, "woof-woof!", 4);
+const cat = new Cat("Cake", GENDER.FEMALE, "meow!", 4);
+const man = new Man("Dominic", "Hello!", 2, 2);
+const woman = new Woman("Dominica", "Hello!", 2, 1);
+
+dog.addFriends("Cake");
+cat.addFriends("Pie", "Selina");
+man.addFriends("Dominica");
+woman.addFriends("Dominic", "Selina");
+
+// ======== OUTPUT ========
+
+const inhabitants = [dog, cat, man, woman];
+
+inhabitants.forEach((inhabitant) => {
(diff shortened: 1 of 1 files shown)


AriadnaKravchuk:
I completed self-checks

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+      this.friends = [];
+      this.attributes = [];
+   }
+
+   addFriends(...friend) {
+      for (let i = 0; i < friends.length; i++) {
+         if (!this.friends.includes(friends[i])) {
+            this.friends.push(friends[i]);
+         }
+      }
```
OleksiyRudenko, line 41: Does this code work? Looks like we've got a bug here.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+   }
+
+   printAttributes() {
+      let friendsNames = [];
+
+      for (let i = 0; i < this.friends.length; i++) {
+         friendsNames.push(this.friends[i]["name"]);
+      }
```
OleksiyRudenko, line 49: A classic use-case for `Array#map`. You need to recognize these as you will use Array methods heavily.

Also you can mix property reference notation `arrayOfObject[index].propName` is an equivalent to `arrayOfObject[index]["propName"]`. It is also better since IDE will be able to check if `.propName` exist, while property names as strings are not essentially testable.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+      print(this.attributes.join("; ") + "; " + friendsNames.join(", "));
```
OleksiyRudenko, line 51: (1) Make class agnostic of environment it works in.
Let the method just return whatever it needs to create and other code to use it (e.g. print).

(2) Separation of concerns. Currently this method does two things: form an output string and prints it.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+      this.legs = legs;
+      this.hands = hands;
+      this.attributes = [
+         this.species,
+         this.name,
+         this.gender,
+         this.saying,
+         this.legs,
+         this.hands
+      ];
```
OleksiyRudenko, line 67: A child class should handle its own properties.
This is the point of encapsulation from OOP and Open/Closed principle from SOLID.
Following these principles help to create maintainable and scalable code.
Imagine, we want to rename one of the shared property.
When we have proper classes design we do this only in the class that owns it. Under current implementation we will need to change code in child classes as well.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+class Animal extends Inhabitant {
+   constructor(species, name, gender, saying, paws) {
+      super(species, name, gender, saying);
+      this.paws = paws;
+      this.attributes = [
+         this.species,
+         this.name,
+         this.gender,
+         this.saying,
+         this.paws
```
OleksiyRudenko, line 80: Code repetition.
Besides, let base class handle properties it owns.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+const dog = new Dog("Pie", GENDER.MALE, 4);
+const cat = new Cat("Cake", GENDER.FEMALE, 4);
+const man = new Man("Dominic", "Hello!", 2, 2);
+const woman = new Woman("Dominica", "Hello!", 2, 1);
+
+dog.addFriends(cat);
+cat.addFriends(dog);
+man.addFriends(woman);
+woman.addFriends(man);
```
OleksiyRudenko, line 117: Add more than 1 friend to test the correctness of the method.

OleksiyRudenko (changes requested):
@AriadnaKravchuk pretty decent job.
Yet, we need to make it OOP compliant.

AriadnaKravchuk:
Thanks for the review!
I rewrote the for loops to map .
I checked the function of adding friends, in the review code there was a typo in the accepted value. I don't see any more errors.
The attribute output has been rewritten with an additional method, it returns the attributes. In inherited classes, it is rewritten with the addition of new attributes.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,124 @@
...
+      this.name = name;
+      this.gender = gender;
+      this.saying = saying;
+      this.friends = [];
+      this.attributes = [
+         this.species,
+         this.name,
+         this.gender,
+         this.saying
+      ];
```
OleksiyRudenko, line 38: All child classes inherit this property. It is never updated and contains only root class properties. So not much use for it outside the root class.
You can safely form the array of attributes in `Inhabitant#getAttributes` method.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,124 @@
...
+   getFriendsNames() {
+      return this.friends.map(({name}) => name);
+   }
+
+   getAttributes() {
+      return this.attributes;
+   }
+
+   printAttributes() {
+      print(this.getAttributes().join("; ") + "; " + this.getFriendsNames().join(", "));
```
OleksiyRudenko, line 56: Try this and take some learning:
```
[
 this.getAttributes(),
 this.getFriendsNames().join(", "),
].join("; ");
```
You will need to recognize patterns for arrays and `Array` methods often. Worth mastering this kung-fu.
Feel free asking questions.

Also the previous comment is still valid. Need to address this.
> (1) Make class agnostic of environment it works in.
> Let the method just return whatever it needs to create and other code to use it (e.g. print).

Let the method return a string representing an object and use print in line 123.

submissions/AriadnaKravchuk/opp-tiny-js-world/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+class Man extends Human {
+   constructor(name, saying, legs, hands) {
+      super(SPECIES.HUMAN, name, GENDER.MALE, saying, legs, hands);
+   }
+}
+
+class Woman extends Human {
+   constructor(name, saying, legs, hands) {
+      super(SPECIES.HUMAN, name, GENDER.FEMALE, saying, legs, hands);
```
OleksiyRudenko, line 103: You can use `SPECIES.HUMAN` in `Human` class and `Human` class' constructor doesn't need species as a parameter. Thus no need to specify species in both `Man` and `Woman` classes.

OleksiyRudenko (changes requested):
@AriadnaKravchuk great improvement!
Let's polish it.

AriadnaKravchuk:
The comments have been corrected

PR #577, opened 2022-09-15
OOP exercise (refactoring TJSW)

MosQuitO404:
### OOP exercise

[Demo](https://mosquito404.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/MosQuitO404/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/MosQuitO404/oop-exercise/index.js
+++ b/submissions/MosQuitO404/oop-exercise/index.js
@@ -16,11 +16,10 @@ class Inhabitant {
       this.gender = gender;
       this.legs = legs;
       this.saying = saying;
-      this.properties = ['species', 'name', 'gender', 'legs', 'saying'];
    }
    
    get showProperties() {
-      return this.properties.map(property => this[property]).join('; ');
+      return ['species', 'name', 'gender', 'legs', 'saying'].map(property => this[property]).join('; ');
    }
 }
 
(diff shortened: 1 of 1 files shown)


MosQuitO404:
@A-Ostrovnyy Hi! I have completed self-checks and fixed code accordingly.

submissions/MosQuitO404/oop-exercise/index.js
```diff
@@ -0,0 +1,90 @@
...
+   constructor(name, gender) {
+      super('Cat', name, gender, 'Meow-meow!');
+   }
+}
+
+class Human extends Inhabitant {
+   constructor(name, gender, saying) {
+      super('Human', name, gender, 2, saying);
+      this.hands = 2;
+      this.properties = ['species', 'name', 'gender', 'legs', 'hands', 'saying'];
```
OleksiyRudenko, line 49: A child class shouldn't handle base class' properties unless there are good reasons to break encapsulation principle.
There is no good reasons for this in this task.
Each class should restrict properties handling to the properties they own.
Look into method overloading and calling inherited versions of methods via `super` (original project docs offer links to relevant materials).

OleksiyRudenko (changes requested):
@MosQuitO404 very well done.
However we need to make the code OOP/SOLID compliant.

submissions/MosQuitO404/oop-exercise/index.js
MosQuitO404: Done!

submissions/MosQuitO404/oop-exercise/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+class Human extends Inhabitant {
+   constructor(name, gender, hands, saying) {
+      super('Human', name, gender, 2, saying);
+      this.hands = hands;
+   }
+
+   get showProperties() {
+      return ['species', 'name', 'gender', 'legs', 'hands', 'saying'].map(property => this[property]).join('; ');
+   };
```
OleksiyRudenko, line 52: > Look into method overloading and **calling inherited versions of methods via `super`** (original project docs offer links to relevant materials).

Take whatever base class creates and append values that this class owns.

This way child class implementation will not depend on base class.

Imagine we add a shared property that base class would handle. In current implementation you'd need to change code in two classes. OOP helps to avoid this and keep code DRY.

OleksiyRudenko (changes requested):
@MosQuitO404 almost there

submissions/MosQuitO404/oop-exercise/index.js
MosQuitO404: @OleksiyRudenko Maybe like this?

OleksiyRudenko (approved):
@MosQuitO404 great job done! Nice and tidy.
P.S. Having classes for men and women might be a bit excessive. Cats and dogs that were not allowed a privilege of having gender-sepcific classes may feel discriminated.

PR #589, opened 2022-09-16
js-dom

merowing:
### js-dom

[Demo](https://merowing.github.io/js-dom/) |
[Code base](https://github.com/merowing/js-dom)

Please, review.


merowing:
- changed for loop on reduce, let to const etc.
- remove .style methods
- check html and css on validator

PR #594, opened 2022-09-17
Classic Frogger Game

YuraZagor:
### Classic Frogger Game

[Demo](https://yurazagor.github.io/frontend-nanodegree-arcade-game) |
[Code base](https://github.com/YuraZagor/frontend-nanodegree-arcade-game)

Please, review.

### DIFF
--- a/submissions/YuraZagor/Frogger/app.js
+++ b/submissions/YuraZagor/Frogger/app.js
@@ -41,13 +41,14 @@ Enemy.prototype.update = function(dt){
     };
     this.checkColision()
 };
-Enemy.prototype.checkColision = function(){   
+Enemy.prototype.checkColision = function(){ 
+    if (this.y === player.y) {
 
-    if ((this.x - this.player.x < collisionOffset ) && (this.player.x - this.x < collisionOffset )) {
-        this.player.x = playerStartX;
-        this.player.y = playerStartY;
-    };    
-
+        if ((this.x - player.x < collisionOffset ) && (player.x - this.x < collisionOffset )) {
+            player.x = playerStartX;
+            player.y = playerStartY;
+        };    
+    }; 
 };
 
 function addBug() {
--- a/submissions/YuraZagor/Frogger/app.js
+++ b/submissions/YuraZagor/Frogger/app.js
@@ -1,18 +1,21 @@
-let allEnemies = []
-let allSpeeds = [200, 250, 350]
-let rangeY = [73, 156, 239]
+let allEnemies = [];
+let allSpeeds = [200, 250, 350];
+let rangeY = [73, 156, 239];
 
 function Player(x, y, sprite) {
     this.x = x;
     this.y = y;
     this.sprite=sprite;
 };
+
 Player.prototype.render = function() {
     ctx.drawImage(Resources.get(this.sprite), this.x, this.y) 
 };
+
 Player.prototype.update = function() {
 };
-let player = new Player ( 202, 405, 'images/char-boy.png' )
+
+let player = new Player ( 202, 405, 'images/char-boy.png' );
 
 function Enemy(x, y, sprite, speedX) {
     Player.call(this, x, y, sprite);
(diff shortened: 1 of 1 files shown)


YuraZagor:
self-check done

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,92 @@
+const allEnemies = [];
+const allSpeeds = [200, 250, 350];
+const rangeY = [73, 156, 239];
```
OleksiyRudenko, line 3: What are these magic numbers?
As a peer developer I am tasked to add another Y. How do I calculate it? Ideally I want to add a row number and let computer do what they were invented to do - calculations.
Also true for other numbers representing pixels across the entire codebase.

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,92 @@
...
+Enemy.prototype.update = function(dt){
+    this.x += this.speedX * dt;
+    if (this.x > fieldWidth){
+        this.x = -55;
+        this.speedX = allSpeeds[Math.floor(Math.random()*3)];
+    };
+    if (this.y === player.y){
+        if ((this.x - player.x < 40 ) && (player.x - this.x < 40 )) {
+            player.x = playerStartX;
+            player.y = playerStartY;
```
OleksiyRudenko, line 41: Item 4.iii of the requirements

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,92 @@
...
+    };
+};
+
+function addBug() {
+    rangeY.forEach ( (levelMark)=>  allEnemies
+    .push( new Enemy(0, levelMark, 'images/enemy-bug.png', allSpeeds[Math.floor(Math.random()*3)] )) 
+    )
+};
+
+player.handleInput = function (direction) {
```
OleksiyRudenko, line 52: One of the purposes of OOP is that class defines all behaviors of the class.
Please explain why this approach was taken.

OleksiyRudenko (changes requested):
@YuraZagor good start. Needs improvements and double-check against requirements.

submissions/YuraZagor/Frogger/app.js
YuraZagor: Done

submissions/YuraZagor/Frogger/app.js
OleksiyRudenko: > this kind of behavior is typical only to player object
That's correct. Also true is that any method on `Player` prototype is typical to player object.

`player.propertyName = function ()` worked, as expected. We still can have functions as object properties.

But what if we have a multiplayer game? OOP helps to have objects blue-prints and avoid code duplication.

Oh, I see. You based enemies on `Player`.
So in fact `Player` is an "any" character.

The good thing was that you came up with having a base class for shared properties and behaviour.
The improvemnt would be to have both `Enemy` and `Player` class extend such base class with their own specifics.
Are you up to improve your code accordingly?

YuraZagor: yes, that way it makes more sense and now I was able to solve ' player.x = playerStartX' omitting .

 Done

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,109 @@
...
+    };
+    this.checkColision()
+};
+
+Enemy.prototype.checkColision = function(){
+    if (this.y === Player.y){
+        if ((this.x - Player.x < collisionOffset ) && (Player.x - this.x < collisionOffset )) {
+            Player.x = playerStartX;
+            Player.y = playerStartY;
+        };
```
OleksiyRudenko, line 53: Before we had here a reference to an instance of `Player` class stored in a global variable.
Now we directly refer here to the class itself.
Check the difference between class and an instance of a class.
Class is a blue-print. Enemies are intended to interact with specific instance (or instances in multiplayer mode) of `Player` class.
Also demo doesn't seem to work.

OleksiyRudenko (changes requested):
@YuraZagor a bit of wrong direction taken.
Check the difference between class and class instance.
Requirements provide a hint on how to resolve the issue.
Take a step back and try again.

submissions/YuraZagor/Frogger/app.js
YuraZagor: not sure that I fully understand it all, but as I get it I should pass the instance as an attribute, so  that if we change the passed object we can still use the same method. Hope that was it

OleksiyRudenko: > ...pass the instance as an **attribute**...
As an **argument**. Just for the sake of correct use of terminology

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,108 @@
...
+const cellX = 101;
+const cellY = 83;
+const spriteRepositioningY = -10
+const playerStartX = 2*cellX;
+const playerStartY = 5*cellY + spriteRepositioningY ;
+const fieldWidth = 5*cellX;
+const fieldHight = 5*cellY;
+const row1 = cellY + spriteRepositioningY
+const row2 = 2*cellY + spriteRepositioningY
+const row3 = 3*cellY + spriteRepositioningY
```
OleksiyRudenko, line 17: Why mixed style of semicolons usage?

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,108 @@
...
+const rangesY = [row1, row2, row3];
```
OleksiyRudenko, line 18: How about `const rangesY = [1, 2, 3].map(some maths here)`?
Whenever you are tempted to write `var1, var2, var3` it is a clear sign that you need to create an array with some maths.

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,108 @@
...
+};
+Enemy.prototype = Object.create(Character.prototype);
+
+Enemy.prototype.update = function(dt){
+    this.x += this.speedX * dt;
+    if (this.x > fieldWidth){
+        this.x = enemyRestartX;
+        this.speedX = allSpeeds[Math.floor(Math.random()*3)];
+    };
+    this.checkColision(player)
```
OleksiyRudenko, line 42: `checkCollision` doesn't refer to global variable `player`, but now `update` does.
So the problem remains.
Feel free to ask specific questions to the hint from the requirement.
Yet another hint: there is a reason that you pass sprite by reference and do not refer to it directly.

OleksiyRudenko (changes requested):
@YuraZagor nearly there

submissions/YuraZagor/Frogger/app.js
YuraZagor: this is really great and concise, never thought that way. Done

submissions/YuraZagor/Frogger/app.js
YuraZagor: My bad
Actually, I didn't have the habbit to use those before the  course,  and now I try to get used to them but at times I forget in haste

YuraZagor:
self-check done
UAT  done

submissions/YuraZagor/Frogger/app.js
```diff
@@ -0,0 +1,108 @@
...
+            this.player.y = playerStartY;
+        };    
+    }; 
+};
+
+function addBug() {
+    rangesY.forEach ( (levelMark)=>  allEnemies
+    .push( new Enemy(0, levelMark, allSpeeds[Math.floor(Math.random()*3)], player)) 
+    )
+};
```
OleksiyRudenko, line 56: This wasn't a part of the requirements. But try to scale that narrow requirement onto a bigger scale.
A well designed function relies only on parameters it receives and doesn't refer to any global variables.
This way it encapsulates business logic it designated to implement.
In big codebases following this principle will pay.

OleksiyRudenko (approved):
@YuraZagor great job done!

PR #608, opened 2022-09-18
OOP Exercise

daniil-bodryagin:
### OOP Exercise

[Demo](https://daniil-bodryagin.github.io/a-tiny-JS-world/index-oop.html) |
[Code base](https://github.com/daniil-bodryagin/a-tiny-JS-world/blob/master/index-oop.js)

Please, review.

### DIFF
--- a/submissions/daniil-bodryagin/oop-exercise/index-oop.js
+++ b/submissions/daniil-bodryagin/oop-exercise/index-oop.js
@@ -1,0 +1,180 @@
+import { print } from './js/lib.js';
+
+class Creature {
+  constructor ({name, gender, legs}) {
+    this.name = name;
+    this.gender = gender;
+    this.legs = legs;
+    this._friends = [];
+  }
+
+  set friends(listOfFriends) {
+    this._friends = listOfFriends;
+  }
+
+  get friends() {
+    return this._friends.map(({name}) => name).join(', ');
+  }
+
+  toString() {
+    const {constructor: {name: species}, name, gender, legs, friends} = this;
+    return [species, name, gender, legs, friends].join('; ');
+  }
+}
+
+class Human extends Creature {
+  constructor ({name, gender, legs = 2, hands = 2, phrase = 'Hello!'}) {
+    super({name, gender, legs});
+    this.phrase = phrase;
+    this.hands = hands;
+  }
+
+  toString() {
+    const {hands, phrase} = this;
+    const creatureString = super.toString();
+    const [species, name, gender, legs, friends] = creatureString.split('; ');
+    return [species, name, gender, legs, hands, phrase, friends].join('; ');
+  }
+}
+
+class Cat extends Creature {
+  constructor ({name, gender, legs = 4}) {
+    super({name, gender, legs});
+    canMeow(this);
+  }
+
+  toString() {
+    const {phrase} = this;
+    const creatureString = super.toString();
+    const [species, name, gender, legs, friends] = creatureString.split('; ');
+    return [species, name, gender, legs, phrase, friends].join('; ');
+  }
+}
+
+function canMeow(creature) {
+  creature.phrase = 'Meow';
+}
+
+class Dog extends Creature {
+  constructor ({name, gender, legs = 4}) {
+    super({name, gender, legs});
+    canWoof(this);
+  }
+
+  toString() {
+    const {phrase} = this;
+    const creatureString = super.toString();
+    const [species, name, gender, legs, friends] = creatureString.split('; ');
+    return [species, name, gender, legs, phrase, friends].join('; ');
+  }
+}
+
+function canWoof(creature) {
+  creature.phrase = 'Woof-woof';
+}
+
+class Catwoman extends Human {
+  constructor ({name, gender, legs, hands}) {
+    super({name, gender, legs, hands});
+    canMeow(this);
+  }
+}
+
+const inhabitantsConfig = [
+  {
+    species: Dog,
+    name: 'Barbos',
+    gender: 'male',
+    friends: ['Eve', 'Adam']
+  },
+  {
+    species: Dog,
+    name: 'Pirate',
+    gender: 'male',
+    friends: ['Kain', 'Avan', 'Luna']
+  },
+  {
+    species: Dog,
+    name: 'Luna',
+    gender: 'female',
+    friends: ['Eve', 'Adam', 'Abel', 'Avan', 'Pirate']
+  },
+  {
+    species: Cat,
+    name: 'Sonya',
+    gender: 'female',
+    friends: ['Eve']
+  },
+  {
+    species: Cat,
+    name: 'Mashka',
+    gender: 'female',
+    friends: ['Avan']
+  },
+  {
+    species: Cat,
+    name: 'Murka',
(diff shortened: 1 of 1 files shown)


daniil-bodryagin:
Self-check is complete.
Review, please.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
```diff
@@ -0,0 +1,180 @@
...
+  set friends(listOfFriends) {
+    this._friends = listOfFriends;
+  }
+
+  get friends() {
+    return this._friends.map(({name}) => name).join(', ');
+  }
+
+  toString() {
+    const {constructor: {name: species}, name, gender, legs, friends} = this;
```
OleksiyRudenko, line 20: How about setting `species` in constructor? it is definitely a property of a creature that exists not only when we want to learn anything about a creature.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
```diff
@@ -0,0 +1,180 @@
...
+    super({name, gender, legs});
+    this.phrase = phrase;
+    this.hands = hands;
+  }
+
+  toString() {
+    const {hands, phrase} = this;
+    const creatureString = super.toString();
+    const [species, name, gender, legs, friends] = creatureString.split('; ');
+    return [species, name, gender, legs, hands, phrase, friends].join('; ');
```
OleksiyRudenko, line 36: I see what you are doing here.
In OOP paradigm child classes should be unaware of how things are built in any base class.
If we want to add yet another shared property to the base class, we will need to change `toString` implementation in many child classes. We do not want this.
You **do not** need to fix this now.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
```diff
@@ -0,0 +1,180 @@
...
+}
+
+class Cat extends Creature {
+  constructor ({name, gender, legs = 4}) {
+    super({name, gender, legs});
+    canMeow(this);
+  }
+
+  toString() {
+    const {phrase} = this;
```
OleksiyRudenko, line 47: `phrase` is obviously a shared property just as legs and or name or friends.
Why treated differently?

OleksiyRudenko (changes requested):
@daniil-bodryagin good job.
A few things yet to clarify.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: I misunderstood the requirement in "Typical mistakes" section: "1.iv : There is no need to specify species at instantiation yet species are printed". Thought that I must avoid to create property `species` in general. Fixed.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: Now I see that I implemented indeed not good solution.
I wanted to keep initial order of properties, but `hands` are to be inserted into the middle of presentation string. So the main question, how to find the place in this string? Parsing comes to mind as the more common approach but at he same time as very complicated. What would be better to do in such cases?

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: In the `Cat` and `Dog` classes `phrase` is the only property which is absent in base class. In `Human` class they are `hands` and `phrase` respectively.
Why didn't I put this property in the base class? Because every cat must say the same - "Meow" (as it was written in the assignment), and if it would be part of base constructor I would be able to pass different phrases as arguments every time I create new class instance.
Why didn't I put this property in child classes with predefined value? I wanted to do this, but the obstacle is `Catwoman` class. I needed to borrow `phrase` property from `Cat` that's why I created special functions just as canMeow() which sets the phrase value to cats and catwomen. Since humans can say different phrases they have this property in constructor.
Why didn't I use function to implement composition only for catwoman and didn't leave cats and dogs with `phrase` property in costructors? It seemed to me that code would look inconsistent then and that it would be right to use special function for adding phrase property everywhere.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
OleksiyRudenko: Keep a map (list) of properties rather than converted string and use semantics like `insertPropertyAfter(property, target)` or `insertBefore...`
Do string conversion at the very last moment.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
OleksiyRudenko: > Why didn't I put this property in the base class? Because every cat must say the same - "Meow" (as it was written in the assignment), and if it would be part of base constructor I would be able to pass different phrases as arguments every time I create new class instance.

This is solvable by sending a relevant value to the parent's constructor.

> Why didn't I put this property in child classes with predefined value? I wanted to do this, but the obstacle is Catwoman class. I needed to borrow phrase property from Cat that's why I created special functions just as canMeow() which sets the phrase value to cats and catwomen. Since humans can say different phrases they have this property in constructor.

... and that can be `canMeow()` (probably a bit refactored), i.e. `super(.... canMeow()...)` is a valid approach.
Also you may want to look at static class methods to solve the problem with CatWoman's phrase. `Cat.getPhrase()` is even more expressive than `canMeow` from the global context.

`canMeow` is the right approach for compositional pattern (as opposed to hierarchical approach of OOP). That's exactly the catch here.

submissions/daniil-bodryagin/oop-exercise/index-oop.js
OleksiyRudenko: `at instantiation` is when we do `new ClassDefinition`. Whatever is within `constructor` is already the **initialization** phase.

daniil-bodryagin: Now I see. Just was confused

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: Ok, thank you!

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: Added `phrase` to base class, replaced global functions by static class methods. Now the code looks better indeed. Thanks!

submissions/daniil-bodryagin/oop-exercise/index-oop.js
daniil-bodryagin: When I added `phrase` to base class the `toString()` turned into simple `super.toString()` everywhere except in Human class. This is only class I leave as it is.

OleksiyRudenko: @daniil-bodryagin Oh, I missed this one. You do not need to overload a method in child class if it's only function is to call an inherited version of the method. It is inherited by child classes and available out of the box by default.

PR #472, opened 2022-09-07
create js-world

kirill8210:
### a tiny JS world

[Demo](https://kirill8210.github.io/a-tiny-JS-world) |
[Code base](https://github.com/kirill8210/a-tiny-JS-world)

The code is submitted in a dedicated feature branch.

Only code files are submitted.

Please, review.

PR #613, opened 2022-09-19
Js world

kirill8210:
moved index.js file to js-world folder


OleksiyRudenko:
Looks like a duplicate of #472 .
Why do we need this PR?

OleksiyRudenko (approved):
@kirill8210 i see. it is COE

PR #616, opened 2022-09-19
a-tiny-JS-world   v2.0   final

YuraZagor:
### a-tiny-JS-world   v2.0   final

[Demo](https://yurazagor.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/YuraZagor/a-tiny-JS-world)

Please, review.

### DIFF
--- a/submissions/YuraZagor/tjsw/index.js
+++ b/submissions/YuraZagor/tjsw/index.js
@@ -1,0 +1,152 @@
+const inhabitants = [];
+
+let catSay = 'meau';
+let dogSay = 'woufff';
+let manSay = 'Wanna_beer_with_ice-cream!';
+let womanSay = 'Are_U_hungry,_dear?';
+
+let womanName = 'Maria';
+let manName = 'Grumio';
+let heDogName = 'Happy';
+let sheDogName = 'Panda';
+let tomCatName = 'Lucky';
+let kittyName = 'Sweety';
+
+const inputElement = document.createElement('form');
+inputElement.classList.add('input--form');
+document.getElementById('main').appendChild(inputElement);
+inputElement.id = 'form';
+
+inputElement.innerHTML = 
+`
+  <h3>Let's populate this tiny World!!!</h3>
+
+  <section class="addNumber" >
+    <label for="num">I wanna add inhabitants, say </label>
+    <select name="num" id="num">
+      <option value="1">1</option>
+      <option value="2">2</option>
+      <option value="3">3</option>
+      <option value="4">4</option>
+      <option value="5">5</option>
+    </select>
+  </section>
+
+  <section class="species">
+    <label for="species">It will be </label>
+	  <select name="species" id="species">
+      <option value="human">Human</option>
+      <option value="animal">Animal</option>
+	  </select>
+  </section>
+
+  <section class="gender">
+    <label for="gender">boy or girl shall I choose? </label>
+	  <select name="gender" id="gender">
+      <option value="male">male</option>
+      <option value="female">female</option>
+	  </select>
+  </section>
+
+  <section class="naming">
+    <label for="naming">So let's come along with a name  </label>
+    <textarea id="naming" name="naming" rows="1" cols="15" placeholder = 'name'></textarea>
+  </section>
+
+  <section class="saying-selection">
+    <label for="saying">Seems that kinda inhabitants would say</label>
+    <input list="sayings" id="saying" name="saying" placeholder = 'saying'>
+    <datalist id="sayings">
+      <option value=${catSay}></option>
+      <option value=${dogSay}></option>
+      <option value=${manSay}></option>
+      <option value=${womanSay}></option>
+
+    </datalist>
+  </section>
+
+  <section class="friend">
+  <label for="friend">They are best friends with </label>
+	  <select name="friend" id="friend">
+    <option value=${manName}>${manName}</option>
+    <option value=${womanName}>${womanName}</option>
+    <option value=${heDogName}>${heDogName}</option>
+    <option value=${sheDogName}>${sheDogName}</option>
+    <option value=${tomCatName}>${tomCatName}</option>
+    <option value=${kittyName}>${kittyName}</option>
+	  </select>
+  </section>
+
+    
+  <section class="submission">
+	  <input type='submit' value='Create!'>
+  </section>
+  `
+
+const form = document.getElementById('form');
+form.addEventListener('submit', useFormValue);
+function useFormValue(event) {
+  event.preventDefault();
+   
+  num = document.getElementById('num').value;
+  species = document.getElementById('species').value;
+  gender = document.getElementById('gender').value;
+  friend = document.getElementById('friend').value;
+
+  document.getElementById('naming').value ? name = document.getElementById('naming').value : name ='Joy';
+  document.getElementById('saying').value ? saying = document.getElementById('saying').value : saying = "Glory to Ukraine";
+
+  multipleCreation(num, species, gender, name, saying, friend);
+};  
+
+class Inhabitant {
+  constructor(species, gender, name, saying, friend){
+    this.species=species;
+    this.gender=gender;
+    this.name=name;
+    this.saying=saying;
+    this.friend=friend;
+  }
+  depict(){
+    return `${this.species}; ${this.name}; ${this.gender}; ${this.saying}; ${this.friend}`
+  };
+  populate(){
+    inhabitants.push(this);
+    print(this.depict(),'div');
+  };
+};
+
+class Animal extends Inhabitant{
+  constructor( species, gender, name = sheDogName, saying = dogSay, friend = womanName, paws=4, tail=1 ){
+    super(species, gender, name, saying, friend);
+    this.paws=paws;
+    this.tail=tail;
+  };
+  depict(){
+    return `${this.species}; ${this.name}; ${this.gender}; ${this.paws}; ${this.tail}; ${this.saying}; ${this.friend}`
+  };
+};
+
+class Human extends Inhabitant{
+  constructor( species, gender, name = manName, saying = manSay, friend=sheDogName, legs=2, hands=2 ){
+    super(species, gender, name, saying, friend);
+    this.legs=legs;
+    this.hands=hands;
+  };
+  depict(){
+    return `${this.species}; ${this.name}; ${this.gender}; ${this.legs}; ${this.hands}; ${this.saying}; ${this.friend}`
+  };
+};
+
+function creation(species, gender, name, saying, friend, legsPaws, handsTail){
+  species.toLowerCase() === 'human' 
+  ? (new Human(species, gender, name, saying, friend, legsPaws, handsTail)).populate()
+  : (new Animal(species, gender, name, saying, friend, legsPaws, handsTail)).populate()
+};
+
+function multipleCreation(num, species, gender, name, saying, friend, legsPaws, handsTail){
+  for (let i = 0; i<num; i++){
+    creation(species, gender, name, saying, friend, legsPaws, handsTail)
+  };
+  
+};


OleksiyRudenko:
Missing submission of TJSW pre-oop. Is it intended or should we consider this as TSJW pre-OOP?
The review process is different for pre-OOP and OOP Exercise is different.

YuraZagor:
Hello,
I missed that I had to submit  first simple variant, I just left for improvements once I made it 18 dsys ago, so v2 you noticed was the OOP exercise ([734e2a4](https://github.com/kottans/frontend-2022-homeworks/pull/616/commits/734e2a49cd5fbe31acb4809be6425c5d14eee62c)  . Now I commited the simple variant [395e326](https://github.com/kottans/frontend-2022-homeworks/pull/616/commits/395e3264388a8b348909ee7940f05f3338705844)  as well

If needed, I will post them to different directories from different branches

Thanks for your attention, regards

OleksiyRudenko:
One task - one dedicated branch - one dedicated PR

YuraZagor:
Underdstood.  Cleaned the mess here by resetting to v2, the pre-OOP version will be https://github.com/kottans/frontend-2022-homeworks/pull/618

YuraZagor:
I have an idea to go further and make subclasses all the way down to a fully depicted representative - like Man, Woman, Kitty, TomCat, heDog, sheDog, so that they have just this.name=name  left in constructor. Then in creation constructor there be less options, but I will have to add a changes console where U can choose what option U want to change. Not sure if that is needed

OleksiyRudenko:
@YuraZagor find the right balance between generalization and specificity when designing the hierarchy.
Setting defaults via classes is not the only option, especially for the properties that may change in the course of instance life cycle.
Params with deafult values may be an appropriate tool. KISS.

YuraZagor:
I see, will work on that in the evening. Because I am not happy  when all the inhabitants by default say the same

thanks a lot

OleksiyRudenko:
> I see, will work on that in the evening. Because I am not happy when all the inhabitants by default say the same
> 
> thanks a lot

They shouldn't. It depends on what classes define default values for parameters.

YuraZagor:
Updated so that defaults are  used vast to make autocreation easy

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,157 @@
...
+  constructor( species, name, saying, friend, gender, paws, tail ){
+    super(species, name, saying, friend, gender);
+    this.paws=paws;
+    this.tail=tail;
+  };
+  populate(){
+    const { species, name, saying, friend, gender, paws, tail} = this
+    const inhabitantProperties = [species, name, saying, friend, gender, paws, tail]
+    print(inhabitantProperties.join('; '),'div');
+    inhabitants.push(this);
```
OleksiyRudenko, line 55: (1) Repetition of code (see lines 77-81) below. Use method overloading/overriding. Also each class should handle the properties it owns immediately. Use extended `super` API to call inherited method version.
(2) Separation of concerns: this method does at least two distinctive things.
(3) Why would class `Animal` (or any class in hierarchy of inhabitants) populate the world? Please do not implement "mating" mechanisms. Anyway, class shouldn't refer to a global variable (constants containing ytuly constant values are still an exception to this rule).

OleksiyRudenko (changes requested):
@YuraZagor need to untangle business logic and make more OOP compliant.

submissions/YuraZagor/tjsw/index.js
YuraZagor: 1) I am rather weak in this topic, tried my best but not sure I got the idea the right way. Anyways, now it seems better structured, and still works
2) Got it, that really adds to readability and makes code DRYer
3) I believe I shall insist: If someone inhabits a place, he lives in it himself being an inhabitant; if I populate a place, I get other people to live in it and I do not need to live there myself. No mating implyed, it can be done with children, or older people, or  say men only (like Valaam island in USSR)

OleksiyRudenko: > 3. I believe I shall insist: If someone inhabits a place, he lives in it himself being an inhabitant; if I populate a place, I get other people to live in it and I do not need to live there myself. No mating implied, it can be done with children, or older people, or  say men only (like Valaam island in USSR)

Requrement re global variables in class definitions mentioned in Frogger task is still valid. It is not specific to Frogger. It is about OOP. You may be weak in any topic but you need to accumulate knowledge and re-use it as you go.

If you want to learn, ask questions and dive deeper into the topic (once again if needed).

Direct references to entities that exist outside class definitions restrict class code usage. Current implementation enforces peer developers to have a single one container for inhabitants and it has to be named `inhabitants`. OOP helps to be flexible and introduce concepts with minimal effect on code.

Better design would allow us to add containers like cities, communities etc - all populated with the same inhabitants (in different combinations). Reiterate over OOP principles if needed and improve this part.

Code should be extensible and maintainable.

"It works" is not a good argument when we're talking about design. 
Imagine the task "we may have 1 to 4 containers, each containing equal number of bags of rice. Container's max capacity is 4. Write a function to calculate total number of bags". The following fragment also works.  doesn't mean it is good.
```
function countBags(containersCount, bagsPerContainer) {
  if (containersCount=== 1) return bagsPerContainer * 1;
  if (containersCount=== 2) return bagsPerContainer * 2;
  if (containersCount=== 3) return bagsPerContainer * 3;
  if (containersCount=== 4) return bagsPerContainer * 4;
}
```

YuraZagor: added passing a destination array for inhabitants instances.  

Also even I can see the repetitive pattern in my function creation, but I haven't come up with a solution yet

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+];
+
+class Inhabitant {  
+  constructor(species, name, saying, friend, gender, habitat=inhabitants){
+    this.species=species;
+    this.name=name;
+    this.saying = saying
+    this.friend=friend;
+    this.gender=gender;
+    this.habitat = habitat
```
OleksiyRudenko, line 42: For the love of God make your code style consistent.
Looks like Frakestein-style composition of code from various sources.

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+  accountInhabitants(){
+    this.habitat.push(this);
+  }
+  commonProperties(){
+    const { species, name, saying, friend, gender} = this
+    const commonProperties = [species, name, saying, friend, gender]
+    return commonProperties
```
OleksiyRudenko, line 50: why not `return [.....]`? Method name already explains what to expect from the method

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+  commonProperties(){
+    const { species, name, saying, friend, gender} = this
+    const commonProperties = [species, name, saying, friend, gender]
+    return commonProperties    
+  }
+};
+
+class Animal extends Inhabitant{
+  constructor( species, name, saying, friend, gender, paws, tail ){
+    super(species, name, saying, friend, gender);
```
OleksiyRudenko, line 56: What `super` is and what it does?

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+  populate(){
+    const { paws, tail} = this
+    const speciesProperties = [ paws, tail]
+    print(this.commonProperties().concat(speciesProperties).join('; '),'div');
+    this.accountInhabitants()
+  };
+};
+
+class Dog extends Animal{
+  constructor( name=randomSelect(animalNamesArr), friend=randomSelect(inhabitants).name, gender='male' ){
```
OleksiyRudenko, line 69: Not a good place to use random params.
Why? Becase it is not obvious what would happen when we're writing `new Cat()`

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+  };
+};
+class Cat extends Animal{
+  constructor( name=randomSelect(animalNamesArr), friend=randomSelect(inhabitants).name, gender='female' ){
+    super('cat', name, catSay, friend, gender, 4, 1);
+  };
+};
+
+class Human extends Inhabitant{
+  constructor( name=randomSelect(womanNamesArr), friend=randomSelect(inhabitants).name, gender='female', legs=2, hands=2){
```
OleksiyRudenko, line 80: Any good reason to default `gender` to `female`?

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+    name: 'Hanna',
+    saying: 'Smert vorogam',
+    friend: 'Vasil',
+    legs: 2,
+    hands: 2,
+  },
+];
+
+class Inhabitant {  
+  constructor(species, name, saying, friend, gender, habitat=inhabitants){
```
OleksiyRudenko, line 36: If neither child class passes a value for `habitat` why we introduce this parameter at all?
Also class shouldn't be aware of any global variables.
The requirements you learned from Frogger task are still applicable.

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+    this.accountInhabitants()
+  };
+};
+
+function creation(className, name, friend, gender){
+  className.toLowerCase() === 'human' 
+  ? (new Human(name, friend, gender)).populate()
+  : className.toLowerCase() === 'cat' 
+  ? (new Cat(name, friend, gender)).populate()
+  : (new Dog(name, friend, gender)).populate()
```
OleksiyRudenko, line 98: You may pass a class as an argument. It is just a function at the end of the day.
The below is totally legal and fine.
```
class A {
}

function createInstance(classConstructor) { .... }

createInstance(A);

```

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,154 @@
...
+  };
+  populate(){
+    const { legs, hands} = this
+    const speciesProperties = [ legs, hands]
+    print(this.commonProperties().concat(speciesProperties).join('; '),'div');
+    this.accountInhabitants()
+  };
+};
+
+function creation(className, name, friend, gender){
```
OleksiyRudenko, line 93: `creation` is not a verb.

OleksiyRudenko (changes requested):
@YuraZagor good improvements.
More improvements are still needed.

submissions/YuraZagor/tjsw/index.js
YuraZagor: done
thougth it would be more evident what it returns that way, but I see I thought that mainly because destructuring wasn't common to me and even now it seems more clear

submissions/YuraZagor/tjsw/index.js
YuraZagor: it saves me from repetedly asigning arguments as this.args by calling parent-class constructor I believe. Also, there is magic about 'this' involved, - we must call 'super' before using this keyword for activation of 'derived class' property

submissions/YuraZagor/tjsw/index.js
YuraZagor: nope, it could ass well be gender : 'male' + name from menNamesArr   combination, just wanted to awoid name/gender confusions when using uncontrolled creation

submissions/YuraZagor/tjsw/index.js
YuraZagor: totally agreed & changed

submissions/YuraZagor/tjsw/index.js
YuraZagor: agreed, that's why I cheeted and used neutral names suitable for both cats and dogs of any gender.  But if I want to implement the option of creating several inhabitants on one click, and omit then having the same one name, suppose I need some kind of that hack

OleksiyRudenko: The code is written for other people. They may need to change it, maintain it and expand it. Take care of them. You will want other developers to take care of you in these regards.

submissions/YuraZagor/tjsw/index.js
OleksiyRudenko: Not the right place in the code to solve this problem. It is not the class' fault that someone may get confused. It's a developer's responsibility to minimize the confusion.

submissions/YuraZagor/tjsw/index.js
YuraZagor: I see.
Added a habitat for dogs only - dogsVolier, so now this param makes more sence

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+    const speciesProperties = [ paws, tail];
+    print(this.commonProperties().concat(speciesProperties).join('; '),'div');
+    this.accountInhabitants()
+  };
+};
+
+class Dog extends Animal{
+  constructor( name, friend, gender ){
+    super('dog', name, dogSay, friend, gender, 4, 1);
+    this.habitat = dogsVolier
```
OleksiyRudenko, line 81: I am confused. LoC 36-44 tell me that `dogsVolier` contain a list of dogs blueprints...
Here it is assigned as a habitat. So what `dogsVolier` is in plain language and what it contains?
What is the purpose of `habitat` property?

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+};
+
+class Animal extends Inhabitant{
+  constructor( species, name, saying, friend, gender, paws, tail ){
+    super(species, name, saying, friend, gender);
+    this.paws = paws;
+    this.tail = tail;
+  };
+  populate(){
```
OleksiyRudenko, line 70: This method does at least 3 clearly distinctive things:
(1) adds current class' properties to the list of properties of a parent class
(2) prints them
(3) calls `this.accountInhabitants` (no arguments and also says it will add many inhabitant**s**)

Separation of Concerns is still a thing.

Also, this method returns nothing, which means that if a create a similar method in a child class (e.g. add a Cat-specific property) I won't be able to re-use it but rather will be forced to redefine it from the scratch. What a waste of code.

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+  constructor( name, friend, gender, legs = 2, hands = 2){
+    super('human', name, humanSay, friend, gender);
+    this.legs = legs;
+    this.hands = hands;
+  };
+  populate(){
+    const { legs, hands} = this;
+    const speciesProperties = [ legs, hands];
+    print(this.commonProperties().concat(speciesProperties).join('; '),'div');
+    this.accountInhabitants()
```
OleksiyRudenko, line 101: Quite a repetiotion of code. OOP helps to reduce code repetition.
Use methods overloading/overriding. Avoid terminal functions (functions that return nothing thus not extensible).

OleksiyRudenko (changes requested):
@YuraZagor you elaborated a quite complicated (and frankly, quite convoluted) system of inhabitants creation. Yet OOP part requires significant improvements.
I'd propose to focus on mastering OOP and make classes design more OOP and SOLID-compliant.

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+  };
+};
+
+function createOne(className, name, friend, gender){
+  className.toLowerCase() === 'human' 
+  ? (new Human(name = randomSelect(womanNamesArr), friend = randomSelect(inhabitants).name, gender = 'female')).populate()
+  : className.toLowerCase() === 'cat' 
+  ? (new Cat(name = randomSelect(animalNamesArr), friend = randomSelect(inhabitants).name, gender = randomSelect(genderArr))).populate()
+  : (new Dog(name = randomSelect(animalNamesArr), friend = randomSelect(inhabitants).name, gender = randomSelect(genderArr))).populate()
```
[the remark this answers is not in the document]
OleksiyRudenko: Just noticed you ignored [a similar proposal](https://github.com/kottans/frontend-2022-homeworks/pull/616#discussion_r990795527).
Feel free if you have any questions.
I am thinking if it makes sense to offer solutions... Sad.

submissions/YuraZagor/tjsw/index.js
YuraZagor: Sorry, didn't think I have to answer it given I can google the answer. Anyways, as I understand it:
1. super() if a callback call to parent class, which is a function object itself. As an object, it has props, one of them is prototype prop, that itself is an object.
2. when called, it creates a new object, which gets the __proto__ prop pointing to it's prototyping function object (parent) props (methods in object-model of a function are also props). 
3. Creating-a-new-obj+assigning-'this.'   model is located in first parent, and is not duplicated is children, they get a prop 'derived' which instructs the function to look deeper in the prototyping tree for this method. That is how new obj 'is aware' of all props and methods of all it's ancestors, if there's more than one.

Can do the same from non-object, but functional point-of-view, but that would be kinda strange in OOP I guess

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+  gender: 'female',
+  name: 'Guffy',
+  saying: dogSay,
+  friend: randomSelect(inhabitants).name,
+  paws: 4,
+  tail: 1,
+},];
+
+class Inhabitant {  
+  constructor(species, name, saying, friend, gender, habitat = inhabitants){
```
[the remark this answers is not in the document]
YuraZagor: I see that.  I reacted as I could to your advice a bit ago: 
_Current implementation enforces peer developers to have a single one container for inhabitants and it has to be named inhabitants. OOP helps to be flexible and introduce concepts with minimal effect on code.
Better design would allow us to add containers like cities, communities etc - all populated with the same inhabitants (in different combinations). Reiterate over OOP principles if needed and improve this part._ 

Also, I do not see any other way but using global array to fulfil the requirement **OOP Exercise 2.i: Friends list is a list of objects refs to other inhabitants rather than just names (strings)** As I get it, they all need to be pushed into that array, and any of them has to have access to that array. May be I am wrong, but these are my first programs and I've had no expirience to rely on

submissions/YuraZagor/tjsw/index.js
YuraZagor: habitat was introduced to try and solve the task you gave me 2 weeks ago:  _Better design would allow us to add containers like cities, communities etc - all populated with the same inhabitants (in different combinations)._

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+
+function createOne(className, name, friend, gender){
+  className.toLowerCase() === 'human' 
+  ? (new Human(name = randomSelect(womanNamesArr), friend = randomSelect(inhabitants).name, gender = 'female')).populate()
+  : className.toLowerCase() === 'cat' 
+  ? (new Cat(name = randomSelect(animalNamesArr), friend = randomSelect(inhabitants).name, gender = randomSelect(genderArr))).populate()
+  : (new Dog(name = randomSelect(animalNamesArr), friend = randomSelect(inhabitants).name, gender = randomSelect(genderArr))).populate()
+};
+
+function createMany(num, species){
```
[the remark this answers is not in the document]
YuraZagor: acknowledged/changed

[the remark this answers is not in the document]
YuraZagor: acknowledged/changed

submissions/YuraZagor/tjsw/index.js
```diff
@@ -0,0 +1,166 @@
...
+  };  
+};
+
+const vasya = new Human ('Vasya', womanName, 'male' );
+vasya.populate();
+createOne('human', womanName, animalName, 'male' );
+createOne('human', randomSelect(womanNamesArr), animalName, 'female' );
+createOne('dog', animalName, 'male', animalName);
+createOne('cat', randomSelect(animalNamesArr), 'female', animalName);
+createMany(4, 'human' );
```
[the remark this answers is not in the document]
YuraZagor: changed

submissions/YuraZagor/tjsw/index.js
YuraZagor: 1) My bad. 
3) renamed to _registerInItsHabitat_ to better reflect what's going on there. 

Not sure how I can do without 2 functions called. Do you mean I should make separate method for printing, and than call both like: 
populate( ) { printingMethod( ); registerInItsHabitat( ) }; ?  

For the last part I can smell a hint to ...rest parameters, but I'm not sure how they can be bound to this.  Upd. They can, which moved all methods down to class Inhabitant. I like that!

submissions/YuraZagor/tjsw/index.js
YuraZagor: refactored

submissions/YuraZagor/tjsw/index.js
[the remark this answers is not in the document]
YuraZagor: sorry to read that. Actually I reacted - I realised I lack structured knowledge and started learning all over again on a deeper and methodological material than crap videos from u-tube and rough MDN-docs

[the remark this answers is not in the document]
YuraZagor: now I get the idea, see the advantages (all problems with my global inhabitantsArray solved) and will work on it

YuraZagor:
Finally it works with Class passed as and argument
Also cleaned by removing the DOM part and removed extra-arrays, so only OOP and most-needed code is left

PR #618, opened 2022-09-19
tiny world pre-OOP

YuraZagor:
### tiny world pre-OOP

[Demo](https://yurazagor.github.io/tjsw_pre_OOP/) |
[Code base](https://github.com/YuraZagor/tjsw_pre_OOP)

Please, review.

### DIFF
--- a/submissions/YuraZagor/tiny world pre-oop/index.js
+++ b/submissions/YuraZagor/tiny world pre-oop/index.js
@@ -47,4 +47,10 @@ cat_woman.friend = woman.name;
 
 cat_woman.saying = cat.saying;
 
-inhabitants.forEach( inhabitant => print(`(${inhabitant.species}; ${inhabitant.name}; ${inhabitant.gender}; ${inhabitant.legs ? inhabitant.legs : inhabitant.paws}; ${inhabitant.hands ? inhabitant.hands : inhabitant.tail}; ${inhabitant.saying}; ${inhabitant.friend}`),'div');
+for (let inhabitant in inhabitants){
+  const {species, name, gender, legs, paws, hands, tail, saying, friend} = inhabitants[inhabitant]
+
+  let textToPrint = `(${species}; ${name}; ${gender}; ${legs ? legs : paws}; ${hands ? hands : tail}; ${saying}; ${friend}`
+
+  print(textToPrint,'div')
+};
--- a/submissions/YuraZagor/tiny world pre-oop/index.js
+++ b/submissions/YuraZagor/tiny world pre-oop/index.js
@@ -2,8 +2,8 @@ const dog = {
   species: "dog",
   name: "JEff",
   gender: "male",
-  legs: 4,
-  hands: 0,
+  paws: 4,
+  tail: 1,
   saying: "bow-wwou!",
   friend: 'everyone',
 };
@@ -11,8 +11,8 @@ const cat = {
   species: "cat",
   name: "Klizma",
   gender: "female",
-  legs: 4,
-  hands: 0,
+  paws: 4,
+  tail: 1,
   saying: "meaaawurrr!",
   friend: 0,
 };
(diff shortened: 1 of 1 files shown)


A-Ostrovnyy:
@YuraZagor please provide the link on the demo page in the first message.

YuraZagor:
Done. I already did the OOP version of this in old repo, so had to create a new repo for this one

OleksiyRudenko (changes requested):
@YuraZagor double-check yoursefl against all requirements linked in the bot's message.

OleksiyRudenko (changes requested):
@YuraZagor the code is not DRY enough.
See the item 3 on the list posted by bot.

YuraZagor:
Simplified readability of the printing message with destructuring, although not sure was it what was needed.  Also, had to change from forEach to simple for ... in

OleksiyRudenko:
Please copy the entire item 3 from the requirements here.
Ask questions if anything is not clear.
Fix code accordingly.
Please do not expect mentors to check code against checkslists for you and to red-pen every single issue.

YuraZagor:
Code is [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), which means that whenever you see a pattern in your code it should be optimized. Examples:
print(dog); print(cat); etc ... should be refactored employing Array.forEach as the least.
`${obj.legs}; ${obj.name}; etc...` (yes, strings are also code) must be refactored employing appropriate Array methods

OleksiyRudenko:
So the repeating part in your long string is `"; "`.
How could we reduce repetition by using `Array#join`?

submissions/YuraZagor/tiny world pre-oop/index.js
```diff
@@ -49,8 +49,9 @@ cat_woman.saying = cat.saying;
 
 inhabitants.forEach((inhabitant)=>{
   const {species, name, gender, legs, paws, hands, tail, saying, friend} = inhabitant;
-  let textToPrint = `${species}; ${name}; ${gender}; ${legs ? legs : paws}; ${hands ? hands : tail}; ${saying}; ${friend}`;
+  const printArr = [species, name, gender, legs ? legs : paws, hands ? hands : tail, saying, friend]
+
```
OleksiyRudenko, line 53: Why not `.join` in this line? (not a rhetoric question)

OleksiyRudenko: Also `textToPrint` and `printArr` are quite generic.
What would you call concepts this variables represent in the real world?

E.g. inhabitants are technically objects, but we still give variables that represent them names that are connected with the business domain.

YuraZagor: not .join()    because we have conditions (ternary)  and they would not be processed if stringified

As per the namings...  **inhabitantRepresentation**  could be less generic I believe

OleksiyRudenko: inhabitantRepresentation sounds just right for a string data type.
If its content is lest of something then it would be probably inhabitantProperties. Nore the "-s" at the end, this way we denote variable contains multiple things.

YuraZagor: I see this logic, thanks

submissions/YuraZagor/tiny world pre-oop/index.js
```diff
@@ -49,8 +49,9 @@ cat_woman.saying = cat.saying;
...
+  print(printArr.join('; '),'div')
```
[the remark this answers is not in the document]
YuraZagor: got it, thanks

OleksiyRudenko (changes requested):
@YuraZagor please test the code and fix bugs

YuraZagor:
Sorry for wasting your time with broken code, re-wrote the erased ending. Now it seems running ok

OleksiyRudenko (approved):
@YuraZagor great job!
For the OOP Exercise task, note that OOP helps to avoid "workarounds" like `legs ? legs : paws, hands ? hands : tail` by designing classes so that every class handles its own properties without much care about what other classes do.

PR #657, opened 2022-09-23
Create PR after students review and some other fixes

unabyband:
### task Friends App

[Demo](https://unabyband.github.io/friends-app/) |
[Code base](https://github.com/unabyband/friends-app)

Please, review.


unabyband:
I've made fixes after the self-checking. 
I hope my code looks more according to requirements now. 
Though, situation when to use "if...else" instead ternary operators and vice versa stays a bit unclear. 
And check my "chat feature" out, please
It can even print smiles :-)
Thank you

unabyband:
Minor fixes for avoiding stale status of the PR and trying to make my code better

PR #661, opened 2022-09-23
Dom-api-task initial commit to homeworks

dmytro-botnarenko:
### Document Object Model

[Demo](https://dmytro-botnarenko.github.io/dom-api-task) |
[Code base](https://github.com/dmytro-botnarenko/dom-api-task)

Please, review.

### DIFF
--- a/submissions/dmytro-botnarenko/dom-api-task/css/landscape.css
+++ b/submissions/dmytro-botnarenko/dom-api-task/css/landscape.css
@@ -1,0 +1,107 @@
+@media screen and (min-width: 461px) and (max-width: 1024px) and (orientation: landscape) {
+  /* ------------SWITCHER--------------- */
+  .switcher {
+    width: 4%;
+    left: 15px;
+  }
+  /* ------------MAIN--------------- */
+  .main__info {
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+  .main__info__name {
+    width: 100%;
+    text-align: center;
+    padding: 0 0 1% 0;
+    font-size: 2rem;
+  }
+  .main__info__personage {
+    flex-direction: row;
+    justify-content: space-around;
+    width: 60%;
+  }
+  .main__info__figure {
+    width: 40%;
+  }
+  .main__info__personage table {
+    width: 50%;
+  }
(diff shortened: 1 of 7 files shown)


dmytro-botnarenko:
I have completed self-checks and fixed code, including [common mistakes](https://kottans.org/documentation/docs/doc/code-review/#dom-api) and [General Requirements](https://kottans.org/documentation/docs/doc/code-review/#general-requirements).

After collecting feedback from students, the following comments and suggestions were collected:

issue1

1. I simplified the adaptive layout a bit. But the burger menus remained for resolutions up to 1024 pixels, but they are designed for landscape device situations (small screen height), where I think the burger is more convenient.
2. As for the width of the menu on mobile devices - I understand revier's wish, but I tried to focus on a wide range of devices and this menu width is suitable even for very narrow screens.
3. Text sticking to the edge of the screen - fixed.
4. I based my design on the principle of the original document without the JS. That is, the JS comes into play only after active actions from the user, but does not participate in the initial assembly of the document structure. This may not be the most optimal / up-to-date approach, but it may have a place, I believe (and as I've seen in some other students' projects with merged PR's).
5. This number of variables is probably related to the situation I described in the previous paragraph. At least in my situation, I don't know how to do without declaring these variables yet.
6. Yes, I can see that the naming of my classes may be a bit off from the BEM rules, but after reviewing my code, I realized that the names I chose are quite convenient and understandable.

issue2

Yes, this "problem" appears due to the appearance of a vertical scrollbar in some situations. At first I decided to apply body {scrollbar-gutter: both-edges;}, which is a very good solution, but unfortunately this property is not supported by Safari and it does not pass validation yet. I don't think it's efficient to make complex solutions with JS libraries in this project, and I don't want to make the creepy Chrome scrollbar always visible either. Therefore, I leave this situation as it is, as I do not consider it very frequent and critical.

issue3

1. The bug with the floating footer has been fixed.
2. The synchronous change of menu field styles is still included as a "what can be done to improve". I've implemented clearing the style of the main menu item when a character is selected in the bottom strip menu, but I haven't yet figured out how the buttons can be synchronously highlighted.

PR #681, opened 2022-09-26
memory-game

kirill8210:
### memory-game

[Demo](https://kirill8210.github.io/kottans_memory) |
[Code base](https://github.com/kirill8210/kottans_memory)

Please, review.

### DIFF
--- a/submissions/kirill8210/memory-game/index.js
+++ b/submissions/kirill8210/memory-game/index.js
@@ -1,0 +1,84 @@
+import arr from './json.js';
+
+const main = document.querySelector('.main');
+const doubleArray = [...arr, ...arr];
+
+const createCard = (arr) => {
+    const {value, id} = arr;
+    const cards = document.createElement('div');
+    cards.className = 'card';
+    cards.dataset.dataid = `${id}`;
+
+    cards.insertAdjacentHTML('afterbegin', `
+        <div id="${id}" class="flip-container">
+            <div class="flipper">
+                <div class="front">
+                    <div class="front_card">You number</div>
+                </div>
+                <div class="back">
+                    <div class="back_card">${value}</div>
+                </div>
+            </div>
+        </div>
+    `);
+
+    return cards
+};
+
+const renderCards = (arr) => {
+    const card = arr.sort(() => 0.5 - Math.random()).map(createCard);
+    main.append(...card);
+};
+
+renderCards(doubleArray);
+
+const toggleCards = document.querySelectorAll('.flipper');
+const hiddenCards = document.querySelectorAll('.card');
+let ArrayCards = [];
+let flipCards = 0;
+
+const flipCard = ({target}) => {
+    const targetCard = target.closest('.card');
+    if (!target.closest('.flip')){
+        targetCard.classList.add('flip');
+        ArrayCards.push(targetCard);
+    }
+
+    if (ArrayCards.length === 2) {
+        hiddenCard();
+    }
+};
+
+const hiddenCard = () => {
+    let oneCard = ArrayCards[0].dataset.dataid;
+    let twoCard = ArrayCards[1].dataset.dataid;
+    if(oneCard === twoCard){
+        ArrayCards.forEach((card) => card.style.visibility = "hidden");
+        flipCards += 2;
+    } else{
+        notHiddenCard();
+    }
+    ArrayCards = [];
+    fullCards();
+};
+
+const notHiddenCard = () => {
+    setTimeout(() => {
+        for (let i = 0; i < hiddenCards.length; i++) {
+            hiddenCards[i].classList.remove('flip');
+        }
+        ArrayCards = []
+    }, 500)
+};
+
+const fullCards = () => {
+    if (flipCards === 12) {
+        setTimeout(() => {
+            alert('YOU WIN!');
+            document.location.reload();
+            flipCards = 0
+        }, 1200);
+    }
+};
+
+toggleCards.forEach(card => card.addEventListener('click', flipCard));


kirill8210:
Please, review.

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+const toggleCards = document.querySelectorAll('.flipper');
+let countFlipCards = 0;
+let openedCardId = 0;
+let awaitFlips = 0;
+let oneCard;
+let twoCard;
+
+const flipCard = ({target}) => {
+    if (awaitFlips === 1) return;
+    if (openedCardId === 0){
```
artem-trubin, line 44: This can be just `if (openedCardId) { ...` since `0` is falsy.

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+    const card = arr.sort(() => 0.5 - Math.random()).map(createCard);
+    mainField.append(...card);
+};
+
+shuffleCards(doubleNumbers);
+
+const toggleCards = document.querySelectorAll('.flipper');
+let countFlipCards = 0;
+let openedCardId = 0;
+let awaitFlips = 0;
```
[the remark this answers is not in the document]
artem-trubin: Oh, nevermind, this is supposed to tell if there are cards that wait for flipping. If you are not invested in the code as the author is, it is hard to understand what's happening here. Let's improve a little.

First - we can change it to something like `animationIsGoing` to indicate that this is purely animational lock of actions for user.
Second - from what I see in the code, I understand that this is either `1` or `0`, so make it a boolean variable, `true`/`false` is better suited for such occasions.

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+const shuffleCards = (arr) => {
+    const card = arr.sort(() => 0.5 - Math.random()).map(createCard);
+    mainField.append(...card);
+};
+
+shuffleCards(doubleNumbers);
+
+const toggleCards = document.querySelectorAll('.flipper');
+let countFlipCards = 0;
+let openedCardId = 0;
```
artem-trubin, line 37: Look, you have `openedCardId` that stores an ID of an opened card and then you have `oneCard` and `twoCard` variables that store cards themselves. There are 3 variables and it could be changed to 1 variable.

Let's see what functionality you suppose them to have:
1. They should indicate if there is a card open right now or not.
2. They should store what Id exactly an opened card has.
3. If the user opens a second card they should give you an option to compare IDs of an opened card and the newly clicked one.

Now, onto making it more clear.

We have 2 options:
First, we can get rid of the `openedCardId` variable, since from what I see in your code, it can only store a real ID when `oneCard` is not empty. But the problem is, `oneCard` stores the whole card. With ID inside it. So you're just doubling data, no need for that.

Second (and I think this one is better), we can get rid of `openedCardId` and `twoCard` completely. The `twoCard` variable in global scope doesn't make any sence, it is only used inside the `flipCard` function. Which means you can only store information about a single opened card in the global scope. And the opened card ID is stored in the `oneCard` anyway.

So you can just leave a single variable - `openedCard` and make it `null` by default.

When user clicks a card you do `if (openedCard)...`, and if it's empty, open the card and store it in this variable.
When user clicks another card, you just operate with a second card in bounds of the `flipCard` function, no need to store it somewhere. Check if the card is the same one and do nothing it that's the case. If not, check if they're equal. If they are - make them dissapear and place `null` to `openedCard` once again.

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+        if (openedCardId === twoCard.dataset.dataid) {
+            openedCardId = 0;
+            twoCard.classList.add('flip');
+            oneCard.style.visibility = 'hidden';
+            twoCard.style.visibility = 'hidden';
+            setTimeout(() => {
+                awaitFlips = 0;
+            }, 800);
+            countFlipCards += 2;
+        } else{
```
artem-trubin, line 61: Space after `else`

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+                oneCard.classList.remove('flip');
+                twoCard.classList.remove('flip');
+                awaitFlips = 0;
+            }, 500)
+        }
+    }
+    if (countFlipCards === 12) {
+        setTimeout(() => {
+            alert('YOU WIN!');
+            document.location.reload();
```
artem-trubin, line 74: It's not realy good to reload the whole page to reset the game. It's a cheap and fast code-wise way, yeah, but imagine your game was far more complicated with heavy scripts and textures, user would need to reload this whole thing every time. Or maybe user has internet connection problems and cannot download the game a second time, it should not disrupt user's experience.

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+            countFlipCards = 0
```
artem-trubin, line 75: Are you sure this line even works, since you reload the page right before it?

submissions/kirill8210/memory-game/index.js
```diff
@@ -0,0 +1,81 @@
...
+        }, 1200);
+    }
+};
+
+toggleCards.forEach(card => card.addEventListener('click', flipCard));
```
artem-trubin, line 80: You should not create an event listener for every card separetely. They do the same thing, use event delegation to only use 1 event listener for the whole board.

artem-trubin (changes requested):
Hey, nice work, let's improve it.

PR #683, opened 2022-09-26
 Friends App

YuraZagor:
### Friends App

[Demo](https://yurazagor.github.io/friendsApp/) |
[Code base](https://github.com/YuraZagor/friendsApp)

Please, review.


YuraZagor:
Self-check done and I know this variant doesn't correspond the advices about using delegation and about adressing via input-values. I did a couple of other variants and they are a failure - although delegation there works ok, my data array is either 'undefined' or 'not iterable' when I try sorting. So I decided to take a couple days rest from this task and start all over next week

YuraZagor:
Self-check done
UAT done

PR #707, opened 2022-09-28
Friends App

iva-stasia:
### Friends App

[Demo](https://iva-stasia.github.io/friends-app/) |
[Code base](https://github.com/iva-stasia/friends-app)

Students tested my demo. I fixed the bugs that were found.

Please, review.


iva-stasia:
Added a "both" button for the gender filter. Tried to simplify some functions a little. I didn't divide the sorting function into two different ones. It seems to me that it looks quite simple. I might be wrong though.

PR #729, opened 2022-09-30
Friends App

Dovahkiin1991:
### Friends App

[Demo](https://dovahkiin1991.github.io/friends-app/) |
[Code base](https://github.com/Dovahkiin1991/friends-app)

Please, review.

### DIFF
--- a/submissions/Dovahkiin1991/friends-app/app.js
+++ b/submissions/Dovahkiin1991/friends-app/app.js
@@ -1,0 +1,270 @@
+
+//const ITEMS_PER_PAGE = 10; feature pagination
+const usersList = [];
+let usersListTemp = usersList;
+
+const callAPI = async () => {
+    const data = await fetch('https://randomuser.me/api/?results=30&nat=ua');
+    handleErrors(data);
+
+    const { results: users } = await data.json();
+
+    return users;
+};
+
+const handleErrors = (response) => {
+    if (!response.ok) {
+        throw Error(response.statusText);
+    }
+    return response;
+}
+
+async function getUsers() {
+    try {
+        const users = await callAPI();
+        usersList.push(...users);
+        renderUsers(usersList);
+    } catch (e) {
+        console.log(e);
+    }
+}
+
+const mainContainer = document.getElementById('main__content');
+let usersListContainer = document.createElement('div');
+usersListContainer.classList.add('users');
+
+const renderUsers = (usersList) => {
+    let users = usersList.map((user) => createUser(user)).join("");
+    usersListContainer.innerHTML = users;
+    mainContainer.appendChild(usersListContainer);
+}
+
+const createUser = (user) => {
+    return `
+        <div class="users__item">
+            <div class="user">
+                <div class="user-block">
+                    <div class="user-poster">
+                        <img src="${user.picture.medium}" alt="${user.name.first + ' ' + user.name.last}">
+                    </div>
+
+                    <div class="user-headings">
+                        <h5><strong>${user.name.first + ' ' + user.name.last}</strong></h5>
+                        <h6>${user.dob.age} years old</h6>
+                    </div>
+
+                    <div class="user-description">
+                        <p>${user.location.country + ', ' + user.location.city}</p>
+                        <p>
+                            <a href="mailto:${user.email}">${user.email}</a>
+                        </p>
+                        <p>
+                            <a href="tel:${user.phone}">${user.phone}</a>
+                        </p>
+                        <p class="uppercase">${user.gender}</p>
+                    </div>
+
+                    <div class="user-button" data-id="${user.name.last.toLowerCase()}">
+                        View Profile
+                    </div>
+                </div>
+            </div>
+        </div>
+    `;
+}
+
+const renderUser = (user) => {
+    let singleUser = `
+        <div class="user user--single">
+            <div class="user-block">
+                <div class="user-poster">
+                    <img src="${user.picture.medium}" alt="${user.name.first + ' ' + user.name.last}">
+                </div>
+
+                <div class="user-headings">
+                    <h5><strong>${user.name.first + ' ' + user.name.last}</strong></h5>
+                    <h6>${user.dob.age} years old</h6>
+                </div>
+
+                <div class="user-description">
+                    <p><strong>User Location:</strong></p>
+                    <p>${new Date(user.dob.date).toLocaleDateString('en-us', { weekday:"long", year:"numeric", month:"short", day:"numeric"}) }</p>
+                    <p>&nbsp;</p>
+                    <p><strong>User Location:</strong></p>
+                    <p>${user.location.country}</p>
+                    <p>${user.location.state}</p>
+                    <p>${user.location.city}</p>
+                    <p>&nbsp;</p>
+                    <p><strong>User Contacts:</strong></p>
+                    <p>
+                        <a href="mailto:${user.email}">${user.email}</a>
+                    </p>
+                    <p>
+                        <a href="tel:${user.phone}">${user.phone}</a>
+                    </p>
+                    <p>&nbsp;</p>
+                    <p><strong>User Gender:</strong></p>
+                    <p class="uppercase">${user.gender}</p>
+                </div>
+
+                <div class="user-button user-button--close">
+                    Back to Friends
+                </div>
+            </div>
+        </div>
+    `;
+    mainContainer.innerHTML = singleUser;
+}
+
+// filters
+let filter = document.querySelector('.filter__block input');
+
+filter.addEventListener('input', (e) => {
+    const value = e.preventDefault();
+    const searchValue = e.target.value.trim().toLowerCase();
+
+    if (searchValue !== '') {
+        const filteredFriends = usersList.filter((user) => {
+            return user.name.first.toLowerCase().indexOf(searchValue) > -1;
+        });
+        usersListTemp = [...filteredFriends];
+
+    } else {
+        usersListTemp = [...usersList];
+    }
+
+    renderUsers(usersListTemp);
+});
+
+document.querySelectorAll('.filter__block input[type="radio"]').forEach((elem) => {
+    elem.addEventListener("change", function(event) {
+        let checkedValue = event.target.id;
+        let filteredFriends = [];
+
+        switch (event.target.name) {
+            case 'alphabet':
+                uncheckRadio('age');
+
+                filteredFriends = usersListTemp.sort((a, b) => {
+                    const nameA = a.name.first.toUpperCase();
+                    const nameB = b.name.first.toUpperCase();
+                    if (nameA < nameB) {
+                        return -1;
+                    }
+                    if (nameA > nameB) {
+                        return 1;
+                    }
+
+                    return 0;
+                });
+
+                if (event.target.id == 'alphabetZA') {
+                    filteredFriends.reverse();
+                }
+
+                break;
+            case 'gender':
+                if (event.target.id === 'male' || event.target.id === 'female') {
+                    filteredFriends = usersListTemp.filter(friend => friend.gender === event.target.id);
+                    
+                } else {
+                    filteredFriends = [...usersListTemp];
+                }
+
+                break;
+            case 'age':
+                uncheckRadio('alphabet');
+ 
+                filteredFriends = usersListTemp.sort((a, b) => a.dob.age - b.dob.age);
+
+                if (event.target.id == 'ageHighest') {
+                    filteredFriends.reverse();
+                }
+
+                break;
+            default:
+                filteredFriends = [...usersList];
+        }
+
+        renderUsers(filteredFriends);
+    });
+});
+
+// reset
+let filterReset = document.querySelector('.filter__block button');
+filterReset.addEventListener('click', (e) => {
+    const value = e.preventDefault();
+    closeUser();
+    uncheckRadio('age');
+    uncheckRadio('alphabet');
+    filter.value = '';
+
+    renderUsers(usersList);
+});
+
+//close user
+const closeUser = () => {
+    let singleUser = document.querySelector('.user--single');
+    
+    if (singleUser) {
+        singleUser.parentNode.removeChild(singleUser);
+    }
+
+    window.location.hash = '';
+}
+
+//read more
+mainContainer.addEventListener('click', (e) => {
+    const value = e.preventDefault();
+
+    if(e.target.closest('.user-button')) {
+        if (!e.target.closest('.user-button').classList.contains('user-button--close')) {
+            let id = e.target.closest('.user-button').dataset.id;
+            window.location.hash = id;
+        } else {
+            closeUser();
+            renderUsers(usersList);
+        }
+        
+    }
+});
+
+//change hash event
+window.addEventListener('hashchange', () => {
+    if (window.location.hash) {
+        const singleUser = usersList.filter((user) => {
+            return user.name.last.toLowerCase().indexOf(location.hash.split('#')[1]) > -1;
+        });
+        renderUser(singleUser[0]);
+    } else {
+        renderUsers(usersList);
+    }
+}, false);
+
+//helper
+const uncheckRadio = (name) => {
+    let radioList = document.getElementsByName(name);
+
+    radioList.forEach((title) => {
+        title.checked = false;
+    });
+}
+
+function arrayEquals(a, b) {
+    return Array.isArray(a) &&
+      Array.isArray(b) &&
+      a.length === b.length &&
+      a.every((val, index) => val === b[index]);
+}
+
+// toggle theme
+const theme = document.querySelector('.header__toggle');
+const body = document.querySelector('body');
+
+theme.addEventListener('click', (e) => {
+    const value = e.preventDefault();
+    body.classList.toggle('theme-dark');
+});
+
+//init
+getUsers();
--- a/submissions/Dovahkiin1991/friends-app/index.html
+++ b/submissions/Dovahkiin1991/friends-app/index.html
@@ -1,0 +1,102 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Friends App</title>
+    <link rel="stylesheet" href="./assets/css/reset.css">
+    <link rel="stylesheet" href="./assets/css/style.css">
+</head>
+<body>
+    <header class="header">
+        <a href="./">Friends App</a>
+
+        <a href="#" class="header__toggle">Toggle Theme</a>
+    </header>
+
+    <div class="main">
+        <div class="main__sidebar">
+            <div class="main__sidebar--inner">
+                <div class="main__sidebar--block">
+                    <h4>Look for a friend:</h4>
+                </div>
+    
+                <div class="main__sidebar--block">
+                    <h6>Search by name:</h6>
+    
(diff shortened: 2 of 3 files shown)


Dovahkiin1991:
I made self-checks, and fixes filter logic. Please check.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
+
+//const ITEMS_PER_PAGE = 10; feature pagination
```
artem-trubin, line 2: Don't leave commented code in the final submission. And an empty first line too.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+};
+
+const handleErrors = (response) => {
+    if (!response.ok) {
+        throw Error(response.statusText);
+    }
+    return response;
+}
+
+const getUsers = async() => {
```
artem-trubin, line 23: There's usually a space around operators and keywords like `async`.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        console.log(e);
+    }
+}
+
+const getOpt = (name) => {
+    return document.querySelector(`.filter__block input[name="${name}"]:checked`);
+}
+
+const mainContainer = document.getElementById('main__content');
+let usersListContainer = document.createElement('div');
```
artem-trubin, line 38: This can be `const`.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        throw Error(response.statusText);
+    }
+    return response;
+}
+
+const getUsers = async() => {
+    try {
+        const users = await callAPI();
+        usersList.push(...users);
+        renderUsers([]);
```
artem-trubin, line 27: Doesn't `renderUsers` require 2 arguments?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+    return document.querySelector(`.filter__block input[name="${name}"]:checked`);
+}
+
+const mainContainer = document.getElementById('main__content');
+let usersListContainer = document.createElement('div');
+usersListContainer.classList.add('users');
+
+const renderUsers = (pipes, usersFiltered) => {
+    let users = pipes.reduce((list, func) => {
+        if (func.name.startsWith('filter')) { // Filter
```
artem-trubin, line 43: Is there really sense in these `// Filter` and `// Sort` comments when it's literally in the code?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        users = usersFiltered.map((user) => createUser(user)).join("");
+    } else {
+        users = users.map((user) => createUser(user)).join("");
+    }
+    
+    usersListContainer.innerHTML = users;
+    mainContainer.appendChild(usersListContainer);
+}
+
+const createUser = (user) => {
```
artem-trubin, line 64: You can destruct user's info here, to not write `user.` every time.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+const getOpt = (name) => {
+    return document.querySelector(`.filter__block input[name="${name}"]:checked`);
+}
+
+const mainContainer = document.getElementById('main__content');
+let usersListContainer = document.createElement('div');
+usersListContainer.classList.add('users');
+
+const renderUsers = (pipes, usersFiltered) => {
+    let users = pipes.reduce((list, func) => {
```
artem-trubin, line 42: Let's make your functions a little closer to the "single responsibility" concept. Your `renderUsers` function doesn't just _renders_ them. It also processes the whole array with filters, sorts, etc. These actions could be separated into another function that receives an array of users, some parameters to process, and returns a new array which then can be used as a basis to build HTML elements and render them to a page.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        }
+    }, usersList);
+
+    usersFilteredList = users;
+    
+    if (usersFiltered) {
+        users = usersFiltered.map((user) => createUser(user)).join("");
+    } else {
+        users = users.map((user) => createUser(user)).join("");
+    }
```
artem-trubin, line 58: This code doesn't look very dry. Find a way to prepare an array in advance and then write what needs to happen with it once.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+            </div>
+        </div>
+    `;
+    mainContainer.innerHTML = singleUser;
+}
+
+// filters
+let filter = document.querySelector('.filter__block input');
+nameInput.addEventListener('input', (e) => {
+    const value = e.preventDefault();
```
artem-trubin, line 144: What is this `value`?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+    
+    if (getOpt('alphabet')) {
+        pipes.push(sortByAlphabet);
+    }
+    if (getOpt('age')) {
+        pipes.push(sortByAge);
+    }
+    if (getOpt('gender')) {
+        pipes.push(filterByGender);
+    }
```
artem-trubin, line 155: This approach is pretty strange. You're working not with data, but with HTML state of the page. And, you're also reaching out to DOM multiple times every time someone writes a letter in the search input. While you could've just selected these html elements once and just checked their status.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+            case 'age':
+                uncheckRadio('alphabet');
+                renderUsers([
+                    filterByName,
+                    filterByGender,
+                    sortByAge
+                ]);
+                break;
+            default:
+                // renderUsers();
```
artem-trubin, line 196: Should users be rendered? Or is it a forgotten comment?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        }
+    });
+});
+
+const filterByName = (user) => {
+    const searchValue = nameInput.value.trim().toLowerCase();
+    return user.name.first.toLowerCase().includes(searchValue);
+}
```
artem-trubin, line 204: When you read someone's code and see `fitlerByName` as a function name, what do you expect from such a function? Usually, such a function would do the following:
1. Get some kind of array that should be filtered as a parameter.
2. Get a name to filter by as a parameter.
3. Make some action to produce a filtered array.
4. Return this filtered array.

Let's see what your function does:
1. Gets user as a parameter.
2. Gets some name out of an HTML element somewhere in the function's scope (that potentially could be absent).
3. Makes some actions with this name.
4. Checks if provided user name includes the produced value.
5. Returns a boolean result.

Do you see where I'm going with this?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+
+const filterByGender = (user) => {
+    const genderRadioOpt = getOpt('gender');
+    if (!genderRadioOpt) { // Pass all if none gender checked
```
artem-trubin, line 208: You can just put a check into the `All` by default. Radiobuttons should always have a default value anyway.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+        }
+    });
+});
+
+const filterByName = (user) => {
+    const searchValue = nameInput.value.trim().toLowerCase();
+    return user.name.first.toLowerCase().includes(searchValue);
+}
+
+const filterByGender = (user) => {
```
artem-trubin, line 206: The comment to `filterByName` applies here as well.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+
+    if (ageRadioOpt) {
+        let isReversed = ageRadioOpt.id === 'ageHighest';
+        return isReversed
+            ? userB.dob.age - userA.dob.age
+            : userA.dob.age - userB.dob.age;
+    }
+
+    return 0;
+}
```
artem-trubin, line 258: First of all, the same comments as to filtering functions.
Secondly, sorting functions essentially work in the same way. They check a and b with some operations and return -1, 1, or 0. Find a way to produce a universal sorting function. It will drastically shorten your code and make it DRY. The trick is then just to pass correct arguments to such a function.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+
+//reset
+let filterReset = document.querySelector('.filter__block button');
```
artem-trubin, line 261: Check [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/reset) out.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+    uncheckRadio('age');
+    uncheckRadio('alphabet');
+    filter.value = '';
+
+    renderUsers([]);
+});
+
+//close user
+const closeUser = () => {
+    let singleUser = document.querySelector('.user--single');
```
artem-trubin, line 278: If you've chosen a coding style - stick to it. Your CSS styles have `__` in them, why not `user__single`?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,339 @@
...
+const theme = document.querySelector('.header__toggle');
+const body = document.querySelector('body');
+
+theme.addEventListener('click', (e) => {
+    const value = e.preventDefault();
+    body.classList.toggle('theme-dark');
+});
+
+//init
+getUsers();
```
artem-trubin, line 339: Add empty lines to the files' end.

artem-trubin (changes requested):
Hi, awesome work! Let's make it better. 

There are some improvements that could be made on a concept level. I saw a lot of reaching out to the dom, a lot of event listeners and a lot of code that is not very DRY. 

Everything could be shortened and written more clearly if you just wrap your inputs in a form and then `querySelect` this form. You can then get any information about the current state of the filters, radiobuttons, etc from this form object by using names. It would also give you 2 things, first of all, you would be able to use one event listener for the whole form and it would enable usage of `input type="reset"` instead of writing a whole function yourself.

Also, check the comments below.

submissions/Dovahkiin1991/friends-app/app.js
Dovahkiin1991: changed to .user.user_single as modificatory

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+    let picture = user.picture.medium;
+    let name = user.name.first + ' ' + user.name.last;
+    let age = user.dob.age;
+    let birthDate = new Date(user.dob.date).toLocaleDateString('en-us', { weekday:"long", year:"numeric", month:"short", day:"numeric"});
+    let country = user.location.country;
+    let state = user.location.state;
+    let city = user.location.city;
+    let email = user.email;
+    let phone = user.phone;
+    let gender = user.gender;
```
artem-trubin, line 96: Why `let`s?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+}
+
+const createUser = (user) => {
+    let picture = user.picture.medium;
+    let name = user.name.first + ' ' + user.name.last;
+    let age = user.dob.age;
+    let location = user.location.country + ', ' + user.location.city;
+    let email = user.email;
+    let phone = user.phone;
+    let gender = user.gender;
```
artem-trubin, line 51: Why `let`s?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+const renderUsers = (filtersObject, usersArray) => {
+    let users = filterUsers(filtersObject, usersArray);
+    users = users.map((user) => createUser(user)).join("");
+    usersListContainer.innerHTML = users;
+    mainContainer.innerHTML = '';
+    mainContainer.appendChild(usersListContainer);
+    window.location.hash = '';
+}
+
+const createUser = (user) => {
```
artem-trubin, line 44: Users aren't created in your work, they are provided by the API. Maybe `createUserCard`?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+const mainContainer = document.getElementById('main__content');
+const usersListContainer = document.createElement('div');
+usersListContainer.classList.add('users');
+
+const renderUsers = (filtersObject, usersArray) => {
+    let users = filterUsers(filtersObject, usersArray);
+    users = users.map((user) => createUser(user)).join("");
```
artem-trubin, line 37: This is why you should consider mostly using constants. `let`s invite mistakes if the project gets more sophisticated.

First you create a variable `users` and you place some array there. Later on, you can mistake it with some global entity `users`, since the name is too general. What actualy is happening is that you store your _filtered_ users here. So why not name it something like `filteredUsers`?

Then you _reassign_ the same variable with another value. It was a filtered array of objects, containing user informations, now it is a string, containting HTML code for the page. In the same variable named `users`. You see the trouble here?

Simple 

```js
const filteredUsers = ...
const userCards = ...
```

would massively improve readability and your future ability to quickly navigate your own code without looking for what's actually inside those variables.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+                    <div class="user-button" data-id="${name.replace(/\s/g, '').toLowerCase()}">
+                        View Profile
+                    </div>
+                </div>
+            </div>
+        </div>
+    `;
+}
+
+const renderUser = (user) => {
```
artem-trubin, line 86: Maybe change it to `renderUserProfile` or something like it. Confusing, looks like the previous function, but slightly different.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+                    <p>
+                        <a href="tel:${phone}">${phone}</a>
+                    </p>
+                    <p>&nbsp;</p>
+                    <p><strong>User Gender:</strong></p>
+                    <p class="uppercase">${gender}</p>
+                </div>
+
+                <div class="user-button user-button_close">
+                    Back to Friends
```
artem-trubin, line 132: Shouldn't this be a link? Let's stay accessible and use the tools provided for us by the browser.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// filters
+const sidebarFilters = document.querySelector('.main__sidebar');
+sidebarFilters.addEventListener("submit", function(event) {
+    event.preventDefault();
+
+    renderUsers(getSidebarOptions(), usersList);
+
+    if (window.screen.width <= 992) {
```
artem-trubin, line 146: `992` is a magic number here. It's a name for random number or specific values that doesn't make sense as is. Declare a global variable like `const MOBILE_VERSION_BREAKPOINT_WIDTH = 992` and use it instead.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+    return filtersObject;
+}
+
+const filterUsers = (filtersObject, usersArray) => {
+    const pipes = [filterByName, filterByGender, sortBy];
+    const filterName = filtersObject.name;
+    const filterGender = filtersObject.gender;
+    const filterSort = filtersObject.filterSort;
+
+    let users = pipes.reduce((list, func) => {
```
artem-trubin, line 167: Why not just do `usersFilteredList = pipes....` here? And then return it. What is the purpose of the `users` variable?

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+
+const filterByGender = (array, filterOption) => {
+    if (filterOption !== 'genderAll') {
+        return array.filter(user => user.gender === filterOption);
+    } else {
+        return array;
+    }
+}
+
+const sortBy = (array, sortValue) => {
```
artem-trubin, line 202: You can make this code even more dry. Both sortings are using the same principle, they get value A, value B, compare it and return -1, 0 or 1. It makes sense for both letters and numbers. So essentially you can have only 1 sorting functions. The trick is to feed it the right arguments. Try to think about how you can do it.

submissions/Dovahkiin1991/friends-app/app.js
```diff
@@ -0,0 +1,304 @@
...
+const theme = document.querySelector('.header__toggle');
+const body = document.querySelector('body');
+
+theme.addEventListener('click', (e) => {
+    e.preventDefault();
+    body.classList.toggle('theme-dark');
+});
+
+//toggle menu
+const menuToggle = document .querySelector('.header__toggle-menu');
```
artem-trubin, line 291: There's a space here :D

submissions/Dovahkiin1991/friends-app/index.html
```diff
@@ -0,0 +1,113 @@
...
+<body>
+    <header class="header">
+        <div class="header__toggle-menu">
+            <span></span>
+            <span></span>
+            <span></span>
+            <span></span>
+        </div>
+
+        <a href="./" class="header__logo">Friends App</a>
```
artem-trubin, line 20: If you do href `#` the page will not reload after the click. We tend to make something like a single-page application here, so it should not reload itself.

submissions/Dovahkiin1991/friends-app/index.html
```diff
@@ -0,0 +1,113 @@
...
+                    </div>
+                </div>
+            </div>
+        </form>
+
+        <div class="main__content" id="main__content"></div>
+    </div>
+    <script src="./assets/js/app.js"></script>
+</body>
+</html>
```
artem-trubin, line 113: There should be an empty line at the end of the file.

artem-trubin (changes requested):
Sorry for the delayed review. Nice work! With some improvements it could be very nice! Please check out the comments below. Also, there are a lot of `let`s here. I didn't tag all of them out, please review them yourself, most of them could and should be changed to `const`.

submissions/Dovahkiin1991/friends-app/app.js
Dovahkiin1991: Lol, but it's still worked with space)

submissions/Dovahkiin1991/friends-app/app.js
Dovahkiin1991: Changed to usersFilteredList = pipes...., but I set usersFilteredList to be as let variable. Cause I'm overriding this value in a few places.

Dovahkiin1991:
Also added empty line at the end of the all files, but don't see it on the GitHub "Files Changed" tab.

